PR #56, opened 2021-12-25
Added Blood Stalk and Grim Swathe Combo features as requested

damolitionn:
Added feature requested in #36 (I think. At least that's what I thought was being requested.) Makes Blood Stalk turn into Gluttony when off-CD and turns into Gibbet/Gallows when used (so all in 1 button). Same with Grim Swathe.

As an aside, I had to build in the GrimSwatheFeature as I figured if people want a 1 button rotation for Blood Stalk -> Gibbet/Gallows they'd want the same for Grim Swathe -> Guillotine. The Feature doesn't differentiate between if the Gluttony is being used for AoE or ST so I had to break it up. So, up to you guys if you want to remove the feature in lieu for the built in option here.

### DIFF
--- a/XIVComboExpanded/Combos/RPR.cs
+++ b/XIVComboExpanded/Combos/RPR.cs
@@ -382,4 +382,57 @@ namespace XIVComboExpandedPlugin.Combos
             return actionID;
         }
     }
+
+    internal class ReaperBloodStalkComboOption : CustomCombo
+    {
+        protected override CustomComboPreset Preset => CustomComboPreset.ReaperBloodSwatheComboFeature;
+
+        protected override uint Invoke(uint actionID, uint lastComboMove, float comboTime, byte level)
+        {
+            var gluttonyCD = GetCooldown(RPR.Gluttony);
+            if (actionID == RPR.BloodStalk)
+            {
+                if (!HasEffect(RPR.Buffs.SoulReaver))
+                {
+                    if ((actionID == RPR.BloodStalk) && !gluttonyCD.IsCooldown && level >= 76)
+                        return RPR.Gluttony;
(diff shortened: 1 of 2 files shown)


Astelteyn:
AOE not soo important) In 95% case i not use AOE)

Astelteyn:
> Added feature requested in #36 (I think. At least that's what I thought was being requested.) Makes Blood Stalk turn into Gluttony when off-CD and turns into Gibbet/Gallows when used (so all in 1 button). Same with Grim Swathe.
> 
> As an aside, I had to build in the GrimSwatheFeature as I figured if people want a 1 button rotation for Blood Stalk -> Gibbet/Gallows they'd want the same for Grim Swathe -> Guillotine. The Feature doesn't differentiate between if the Gluttony is being used for AoE or ST so I had to break it up. So, up to you guys if you want to remove the feature in lieu for the built in option here.

Em, but where?=_= or it only in test version or need modify files and compile.....

damolitionn:
Actually yeah, I don't see my changes. Maybe it wasn't accepted post testing.

Nik-Potokar:
Seems like it was only added into testing branch and it didn't transfer (when the branch got renamed)
 I will fix when I'm home unless you can do another request for main branch merge so that I can do it on mobile 

Sorry about that

Astelteyn:
This is not so urgent, I will wait for the update of the main branch. The main thing is that it will be added)

Astelteyn:
But by the way, I tried to establish a test assembly at this address, but nothing happened, maybe I did something wrong 
https://raw.githubusercontent.com/Nik-Potokar/MyDalamudPlugins/testing/pluginmaster.json

PR #77, opened 2021-12-29
SMN One button combo level sync fixes

Genesis-Nova:
Level sync fix for Searing Light one button combo (need to be level 66 or greater, wasn't working while sub 66)
Also fix for one button combo egi summons, was returning summon titan (I) instead of summon titan II at level 90 and was nonfunctional.
Also fixed one button combo for pre-bahamut level sync, had summon bahamut when you couldn't click it.
Unsure if I did this entirely correctly but it does seem to work fine in-game when testing at sub 66 for searing light, and sub 30/post 30 for carbuncles and egis.


Nik-Potokar:
Seems to be working without any issues so far 
Thank you very much for saving me some time <3

PR #82, opened 2021-12-30
Fell Cleave/Decimate Option + Edit to Inner Chaos Option

damolitionn:
Added the option for the ST/AoE Combos to use their spenders at 50 instead of at overcap with level checks. As well as, adding Chaotic Cyclone to the AoE combo when the Inner Chaos Option is ticked.


Codemned:
Inner release feature already adds Chaotic Cyclone to the aoe combo

damolitionn:
Ahh, when I was testing it I didn't have that ticked seeing as the description says during Inner Release and I wanted to see if the Inner Chaos option worked when using Infuriate.

Issue #86, opened 2021-12-30
[RPR] New enshroud combo falls apart if you do the combo out of order


sauronshotdogs:
If I do Cross reaping first instead of void, it will then try to do Cross reaping again, instead of Lemure's slice, thus making the combo fall apart.

Could you make it so it works, even if you do some of the combo manually?

damolitionn:
That's what I get for throwing a bandaid fix to check for the Enhanced buff. I'll take a look at it in the next hour

sauronshotdogs:
I am just happy have have something. I am so used to using it in the
opposite order of what you did, that I press it manually out of order.

The combo that I referenced in the original post works regardless of order
pressed, so you might be able to steal some code.



Venlig hilsen,
Jacob Jordansen

PR #87, opened 2021-12-31
Fixed Enshroud Combo breaking if started manually

damolitionn:
Should fix the issue in #86. I've tested it starting it on both Void or Cross Reaping and it hasn't broken yet so it should be good now.


sauronshotdogs:
Mind sharing a compile? I don't know how to pull your fix myself

PR #114, opened 2022-01-08
chore: refactor SimpleBardFeature in BRD

ele-starshade:
Another little refactor to make the function more performant, easier to understand and DRY.

Removed un-necessary braces from if conditionals
Refactored duplicated heavy shot cooldown checks into one variable
Refactored duplicated inCombat checks into blocks
Renamed variables to be camelCase instead of all lowercase, to be consistent with other variables around the codebase
Grouped the useIronJaws check into one single variable to ease readability
Moved level check conditionals to be at the start of the conditionals to save function calls on lower level content

All tested, builds and rotation still working as it was prior to the refactor

Thanks! :)


Nik-Potokar:
Thank you!

PR #118, opened 2022-01-09
Fixed lazy fester (single and aoe) not working with phoenix.

Genesis-Nova:
Lazy fester wasn't working with Phoenix, just added checks for that and it seems to be working fine in my small amount of testing.


Nik-Potokar:
That is intended because you are saving your festers for your Bahamut burst phase

Genesis-Nova:
Hmm I see, haven't looked too much into SMN's 'optimal' rotation and I'm just poking around other jobs, guess its pop energy drain during phoenix and use the ogcds during bahamut?

Nik-Potokar:
Yup since the buff will be available when bahamut is ready (unless playing the speed build then everything is diff)

PR #121, opened 2022-01-11
BRD SimpleBard fix and optimization

Codemned:
Corrects wrong song, added pitch perfect spender and moved oGCD's to prioritize song uptime.


goofy-somebody:
When was Mage's Ballad changed to 36s?  still shows to be 43-34-43

Codemned:
I just changed it to 42-36-42 because it'll switch right after the last proc on wanderer/army and it flowed better with automation when I tested it.

It still stays inline with the balance recommended uptimes: https://www.thebalanceffxiv.com/jobs/ranged/bard/basic-guide/#song-order

ele-starshade:
Thanks @Codemned!

PR #127, opened 2022-01-13
Warrior fixes

sauronshotdogs:
Fixed OnslaughtOptionTwo and adjusted description. Now leaves 1 charge below 88.

Moved Orogeny combo up, so it will activate in time with Upheaval. Will still wait for Storm's buff

Attempt at cleaning up my ugly formatting

### DIFF
--- a/XIVSlothCombo/Combos/MCH.cs
+++ b/XIVSlothCombo/Combos/MCH.cs
@@ -119,7 +119,6 @@ namespace XIVSlothComboPlugin.Combos
                     else
                     if (level >= 84)
                     {
-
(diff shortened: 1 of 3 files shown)


Codemned:
It's nice that your trying to learn this and I'd like to give you some feedback, I hope you don't mind.

sauronshotdogs:
Of course!

Codemned:
2 stack option probably needs "if (level >= 62 && level <= 87)" or someone level 88+ would qualify for both arguments bugging the feature.

sauronshotdogs:
Will Try to add that. It works though. I found the order in which the code would appear to be most important, and I wasn't sure how to force the order with arguments

PR #130, opened 2022-01-13
feat: bard simple buffs and radiant

ele-starshade:
Adds a simple buffs and also separate radiant finale feature.

Preliminary release, no optimisation, just does them on cooldown when they are available, but after the songs logic.


Nik-Potokar:
Thank you!

PR #149, opened 2022-01-18
BRD and MCH stuff

sauronshotdogs:
-Fxed BRD interrupt feature.
-Added MCH Gauss / Ricochet drift prevention feature. Will put on main combo and leave 1 charge of each.
-Added MCH Barrel Stabilizer drift prevention feature. Will add Barrel Stabilizer to main combo is between 5 and 20 heat.


sauronshotdogs:
Is there a way to check for the length of the cast that you're interrupting? Sometimes a bot has several interruptables, but you only want to interrupt the long ones

DurzoThorn:
Is there any chance we could take out the automatic usage of Apex Arrow at 100 gauge from the SimpleBard option and just add it as an optional feature? Would be pretty much perfect if I could hold my 2nd Apex Arrow for 2 min burst.

Issue #141, opened 2022-01-16
[GNB] Many, many changes wanted.


Trilkin:
These are pretty simple:

Add Continuation To Main Combo (if the appropriate buffs are up)
Add Gnashing Fang Combo to Main Combo (ONLY if you've already started the combo to prevent you from accidentally wasting it)

Iaotle:
Honestly GNB needs a general rework. 
* You should be able to double weave all of them in your No Mercy windows. 
* Gnashing Fang should easily be combo-able into your main combo.
* OGCD priorities are messed up I think, needs to be looked at.

Trilkin:
> Honestly GNB needs a general rework.
> 
> * You should be able to double weave all of them in your No Mercy windows.
> * Gnashing Fang should easily be combo-able into your main combo.
> * OGCD priorities are messed up I think, needs to be looked at.

As long as Gnashing Fang is a separate option. There are moments where you don't want to immediately use it (P2S limit cut, for example)

Iaotle:
Don't worry the whole point of this plugin is to provide options (see SMN).

damolitionn:
Added the features you requested and also redid most of the things in GNB because it was done like someone read a TLDR of GNB without actually playing the class.

I wish I could have done things like add Bloodfest into the main combo and throw in No Mercy into the main combo but it messes up the opener. I guess it's fine if you're not optimising so you could start a fight with Bloodfest into No Mercy and then into the burst but it just feels wrong to do that, it already felt weird adding Rough Divide outside of buff windows. I guess if people request it I can add it as another modular option into the main combo.

Trilkin:
You could add it as an option for the same reason bard has similar options, I'd figure: so they get used on cooldown. If you're trying to optimize, you're always going to do your opener manually. I'll test the changes out in savages once they get pushed out and reply here after a few runs. If all seems good, I'll close this ticket.

damolitionn:
Yeah I can throw it in after you're satisfied with the base updates first. I've always been one to add things that I personally might use so it feels off to me to include those options but I can see it being used if you're super lazy and you don't worry too much about losing usage because of uptime.

Trilkin:
Ye, I personally wouldn't use it. I don't use most of the bard or summoner options for the same reason. Too much automation actually ends up hurting rather than helping.

Trilkin:
[...]

PR #155, opened 2022-01-22
GNB Overhaul

damolitionn:
Complete overhaul of GNB including features asked for in #141 (who even did the original because that was rough).

Features added:
 - Added Gnashing Fang, Continuation, Double Down, Bow Shock, Sonic Break, Blasting Zone/Danger Zone, Rough Divide into main combo. Gnashing Fang will need to be started manually (as requested in the #141) in order for users to have more control during unfortunate phase change timing. These abilities are modular for increased choice in the main combo
 - Overcap features for both ST and AoE
 - No Mercy Rotation on No Mercy: Added the full Gnashing Fang/DD/oGCDs/Continuation burst into No Mercy.


griminir:
when will this update be released ?

PR #196, opened 2022-01-28
[BLM] Update to all-in-one rotation and aoe for level sync

grimgal:
Change Scathe to Fire 3 when MP is maxed, in Umbral Ice and with 3 Umbral Hearts.
Change Scathe to Blizzard 3 when MP is 0 and in Astral Fire.
Change Scathe to Blizzard 3 when MP is below 1600 and in Astral Fire at low levels.
Change Freeze to be cast only when MP is below 10000 in Umbral Ice at low levels since Umbral Hearts aren't granted.

Tested changes locally and should be fine, but feel free to make comments.

### DIFF
--- a/XIVSlothCombo/Combos/BLM.cs
+++ b/XIVSlothCombo/Combos/BLM.cs
@@ -265,10 +265,19 @@ namespace XIVSlothComboPlugin.Combos
                             return BLM.HighBlizzardII;
                     }
                 }
-                if (IsEnabled(CustomComboPreset.BlackAoEComboFeature) && level >= BLM.Levels.Blizzard4)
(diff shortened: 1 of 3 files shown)


Nik-Potokar:
Will test when I'm home but I would prefer if this would be a seperate toggle since it seems to automate most of the rotation and that would interfere with current intended options that let the user manually swap fire/ice stances and it adjsuts from there on.

grimgal:
Yeah, makes sense; I've added the BlackAspectSwap toggle for automating this. Also, I added some fixes for low levels before you get umbral hearts to the one button AoE that should make the AoE combo work for all low levels.

PR #238, opened 2022-02-05
sam fix

pynom:
- Fixes bug with Shoha being added to laijutsu while YukikazeCombo was active
- Adds Kaiten level check
- Allows player to better see Tsubame stacks
- Fixes st overcap bug always stuck on 75 kenki due to incorrect numbers
- Fixes aoe overcap bug active on incorrect feature due to invalid conditions


Iaotle:
Can you merge both PRs into one please?

PR #254, opened 2022-02-11
Adding a sub feature to simple bard - Simple RagingJaws

augporto:
### What it does

- Add the ability to toggle a feature that snapshot buffed up dots in the end of the raging strike buff window when using SimpleBard.

### General improvements

- Raises up the priority of sidewinder in relation to Bloodletters, making it drift less.
- Make sure Radiant Finale is only called after Battle Voice, to better align buffs in the burst window.
   - RS, RF and BV should be ending all at the same time now.
- Changing song use timings to better represent what theBalance sugests.
   - WM  < 3s 
   - MB < 12s 
   - AP < 3s or WM is off cooldown
- Added RA, SS and IJ to CanWeave
   - this helped with the drifting some skills had when I analyzed the logs.

### DIFF
--- a/XIVSlothCombo/Combos/BRD.cs
+++ b/XIVSlothCombo/Combos/BRD.cs
@@ -527,6 +527,8 @@ namespace XIVSlothComboPlugin.Combos
         internal static byte subStep = 0;
         internal static bool usedStraightShotReady = false;
 
+        internal delegate bool DotRecast(int value);
+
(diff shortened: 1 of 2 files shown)


ele-starshade:
Thanks @augporto - Just the conflict, if you could poke up the number by one on yours that would be great, then we can get it merged :)

PR #255, opened 2022-02-12
WIP: chore(AST): re-work ast class

ele-starshade:
- [x] Test it!
- [ ] Do others

### DIFF
--- a/XIVSlothCombo/Classes/AST.cs
+++ b/XIVSlothCombo/Classes/AST.cs
@@ -1,4 +1,5 @@
-using System.Collections.Generic;  
+using System.Collections.Generic;
+using Dalamud.Game.ClientState.JobGauge.Types;
 
 namespace XIVSlothComboPlugin.Classes
 {
@@ -179,5 +180,8 @@ namespace XIVSlothComboPlugin.Classes
 
       return level >= requiredLevel;
     }
+
+    public static ASTGauge JobGauge =>
+      Service.ComboCache.GetJobGauge<ASTGauge>();
   }
 }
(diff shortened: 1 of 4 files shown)


Iaotle:
Wow, good job! You're doing a lot of good work.

augporto:
Hello @ele-starshade, @Iaotle  would you guys like some help to replicate this rework on the other jobs ? I really like it 👍

Iaotle:
@augporto I think we need to add the Dictionary to some global or something so we can use it for all jobs, will prevent us from having to make one per job.

Otherwise, I might have time end of next week to work on something. Not sure tbh, I got uni.

Iaotle:
One other thing, maybe we can have a `ReturnIfUnlocked(spell, level)` func in order to simplify the logic. That way it will return 0 if not unlocked, and spell ID if it is.

augporto:
> @augporto I think we need to add the Dictionary to some global or something so we can use it for all jobs, will prevent us from having to make one per job.
> 
> Otherwise, I might have time end of next week to work on something. Not sure tbh, I got uni.

Yea I was thinking about that too, maybe a global class Job that others can use with inheritance to centralize normal functions and shared variables like that action dictionary to all jobs like the unlock checks , to not need to rewrite the same function on all classes again. But initialization would be per Job based , or else we would have a ginormous initialization.

> One other thing, maybe we can have a `ReturnIfUnlocked(spell, level)` func in order to simplify the logic. That way it will return 0 if not unlocked, and spell ID if it is.

That seems good too.

Iaotle:
This should be more of a maintenance thing tbh, I believe the codebase changes here are great if someone wants to pick it up @k-kz

k-kz:
I've created a 'placeholder' version of this, disconnected from AST on a branch of my fork for posterity, but at this point I think we've superseded the need for a `Levels` dictionary (we no longer need to store level constants at all, when using `LevelChecked()`), so I'll be closing this.

https://github.com/k-kz/XIVSlothCombo/blob/dictionary/XIVSlothCombo/ActionDictionary.cs

PR #261, opened 2022-02-13
[BRD] oGcds in the opener

augporto:
Greetings 🖖 

### What this PR does

- Smooths out the oGcds in the opener , basically PitchPerfect and Bloodletter
    - It is for now more responsive, and consistent with weaving between the opener GCDs.
    - Got rid (while I tested) of those triple weaving issues that happened sometimes.

### Test sample

- With RA proc before barrage (I think this was where most of the triple weaving happened for me):

image

   - this one here is where a PP happens together with a sidewinder in the same situation as above, instead of the BL

image

- Without RA proc before barrage:

image

this was a lot of fun to tackle , frustrating sometimes lol, but fun nonetheless.

Tyvm for all the work you all do on this plugin!!

### DIFF
--- a/XIVSlothCombo/Combos/BRD.cs
+++ b/XIVSlothCombo/Combos/BRD.cs
@@ -534,7 +534,9 @@ namespace XIVSlothComboPlugin.Combos
         internal static bool openerFinished = false;
         internal static byte step = 0;
         internal static byte subStep = 0;
+
         internal static bool usedStraightShotReady = false;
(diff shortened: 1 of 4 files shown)


sauronshotdogs:
Got some fflogs for the new version?

augporto:
https://xivanalysis.com/fflogs/cn7V94BzJbtANWD1

here is the analysis of the last ones I used , the last 3 of them are the ones I printed imgs of, I think the second one there even have a triple weaving happening.

PR #264, opened 2022-02-15
[BRD] Optimized Army Paeon pooling protection

augporto:
### Enhancements

- An optimized protection for the pooling feature for BL and now Sidewinder.
    - Makes both of them to only be called, when under WM, if all the buffs are active.
        -  Minimizing wasting them before getting full power.
- BV and RF GCD drift protection
    - Making RS only be called when BV is off CD or 2 GCDs from being off CD, better aligning all buffs.

### Minor things

- Added 3 STs gcds to canWeave of DNC on SimpleDancer.
    - Fountain , Fountainlfall and SaberDance.


augporto:
Found out that the SaberDance skill on simple dancer was not being called when I was playing today , there was an IsOffCooldown check that was preventing it, removing it fixed it, added it to this PR as well.

PR #270, opened 2022-02-20
[RDM] Specialized opener and some other goodies

augporto:
Greetings,

Was leveling RDM this weekend and ended up doing a couple of things for SimpleRedMage and was pretty satisfied with the results.

### What is in this PR

- lv90 specialized opener feature (inspired by stars work on brd opener).
    - Did this with trials and raids in mind, disable for openworld is recommended.
    - Its disabled on lower lvs automatically so its safe to do roulletes with it on.  
- Feature to auto use accel and swiftcast to fish for ver procs when necessary , minimizing the use of jolts.
    - You can opt to only use accels and save switcast if you wish to do so 
- Added some protections for melee combos , aoes and sts, to not break as easily, the combos should take priority over other things when they are happening.
   - Should be fine for now, at least in the situations I encountered, could use some feedback. 

### General 

- Thought the the calcbestaction for fleche and countresixte was a bit excessive, so just simplified it.

### Known issues

   - None that I know off.

### To do list ?

   - Maybe automate buffs on simple red mage, maybe... 

Have a nice rest of weekend/week, cheers. o/

### DIFF
--- a/XIVSlothCombo/Combos/RDM.cs
+++ b/XIVSlothCombo/Combos/RDM.cs
@@ -91,13 +91,14 @@ namespace XIVSlothComboPlugin.Combos
                 if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                     return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                if (level >= RDM.Levels.ContreSixte)
+                if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                     return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                if (level >= RDM.Levels.Fleche)
+                if (level >= RDM.Levels.Fleche && level >= RDM.Levels.ContreSixte)
                     return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                return RDM.Fleche;
+                if (level >= RDM.Levels.Fleche && IsOffCooldown(RDM.Fleche))
+                    return RDM.Fleche;
             }
             if (actionID == RDM.Veraero2)
             {
@@ -136,13 +137,14 @@ namespace XIVSlothComboPlugin.Combos
                     if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    if (level >= RDM.Levels.ContreSixte)
+                    if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    if (level >= RDM.Levels.Fleche)
+                    if (level >= RDM.Levels.Fleche && level >= RDM.Levels.ContreSixte)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    return RDM.Fleche;
+                    if (level >= RDM.Levels.Fleche && IsOffCooldown(RDM.Fleche))
+                        return RDM.Fleche;
                 }
                 if (IsEnabled(CustomComboPreset.RedMageEngagementFeature) && actionIDCD.IsCooldown && engagementCD.CooldownRemaining < 35 && InMeleeRange(true))
                 {
@@ -208,13 +210,14 @@ namespace XIVSlothComboPlugin.Combos
                     if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    if (level >= RDM.Levels.ContreSixte)
+                    if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    if (level >= RDM.Levels.Fleche)
+                    if (level >= RDM.Levels.Fleche && level >= RDM.Levels.ContreSixte)
                         return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    return RDM.Fleche;
+                    if (level >= RDM.Levels.Fleche && IsOffCooldown(RDM.Fleche))
+                        return RDM.Fleche;
                 }
                 if (IsEnabled(CustomComboPreset.RedmageResolutionFinisher))
                 {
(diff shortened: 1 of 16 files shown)


XIVSlothCombo/Combos/RDM.cs
```diff
@@ -643,56 +656,258 @@ const int
...
+        internal static bool openerFinished = false;
+        internal static byte step = 0;
+
         protected override uint Invoke(uint actionID, uint lastComboMove, float comboTime, byte level)
 
         {
-            if (actionID == RDM.Veraero || actionID == RDM.Verthunder || actionID == RDM.Verstone || actionID == RDM.Verfire || actionID == RDM.Riposte || actionID == RDM.Zwerchhau || actionID == RDM.Redoublement)
+            if (actionID == RDM.Jolt || actionID == RDM.Veraero || actionID == RDM.Verthunder || 
+                actionID == RDM.Verstone || actionID == RDM.Verfire || actionID == RDM.Riposte || 
+                actionID == RDM.Zwerchhau || actionID == RDM.Redoublement)
```
Iaotle, line 669: I think in C# you can use the ```is A or B or C``` macro instead of chaining ```== || ==```

XIVSlothCombo/Combos/RDM.cs
```diff
@@ -643,56 +656,258 @@ const int
...
+                            if (lastComboMove == RDM.Scorch) step++;
+                            else return RDM.Scorch;
+                        }
+
+                        if (step == 18)
+                        {
+                            if (lastComboMove == RDM.Resolution) step++;
+                            else return RDM.Resolution;
+                        }
+
```
Iaotle, line 845: I see what you're doing here, but honestly should this be less of a multistep combo you think? It might be a bit on the automation heavy side, this is an 18 move combo... idk

XIVSlothCombo/Combos/RDM.cs
```diff
@@ -843,36 +1058,47 @@ const int
...
                 int whiteThreshold = black + IMBALANCE_DIFF_MAX;
-                var actionIDCD = GetCooldown(actionID);
-                if (IsEnabled(CustomComboPreset.RedMageOgcdComboOnCombos) && actionIDCD.IsCooldown && IsOffCooldown(RDM.Fleche) || IsEnabled(CustomComboPreset.RedMageOgcdComboOnCombos) && actionIDCD.IsCooldown && IsOffCooldown(RDM.ContreSixte))
-                {
-                    if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
-                        return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
 
-                    if (level >= RDM.Levels.ContreSixte && level <= RDM.Levels.Fleche)
-                        return CalcBestAction(actionID, RDM.ContreSixte, RDM.Fleche);
+                var canWeave = CanWeave(actionID);
```
Iaotle, line 1062: If you use it once, you shouldn't make a var for it IMO.

XIVSlothCombo/Combos/RDM.cs
```diff
@@ -891,17 +1117,18 @@ const int
 
                     return RDM.Verflare;
                 }
-                if (dualcastBuff || accelBuff || swiftcastBuff || HasEffect(RDM.Buffs.Chainspell) || level <= 18)
+                if (dualcastBuff || accelBuff || swiftcastBuff || HasEffect(RDM.Buffs.Chainspell) )
                     return OriginalHook(RDM.Impact);
 
-                if (level <= 18)
-                    return RDM.Verthunder2;
+                if (level < RDM.Levels.Verthunder2)
```
Iaotle, line 1123: I saw a level map somewhere by the way, we should implement that solution-wide.

XIVSlothCombo/CustomComboPreset.cs
```diff
@@ -1033,6 +1033,19 @@ public enum CustomComboPreset
         [CustomComboInfo("oGCD Feature Everywhere", "Adds oGCD Feature to all other combos", RDM.JobID)]
         RedMageOgcdComboOnCombos = 13017,
 
+        [ParentCombo(SimpleRedMage)]
+        [CustomComboInfo("Simple Proc Fishing","Add Acceleration and Swiftcast to Simple Red Mage for proc fishing when necessary.", RDM.JobID)]
+        SimpleRedMageFishing = 13018,
+
+        [ParentCombo(SimpleRedMageFishing)]
+        [CustomComboInfo("Save swiftcast", "Only use Acceleration to fish.", RDM.JobID)]
```
Iaotle, line 1041: Needs a better description I think.

XIVSlothCombo/CustomComboPreset.cs
```diff
@@ -1033,6 +1033,19 @@ public enum CustomComboPreset
         [CustomComboInfo("oGCD Feature Everywhere", "Adds oGCD Feature to all other combos", RDM.JobID)]
         RedMageOgcdComboOnCombos = 13017,
 
+        [ParentCombo(SimpleRedMage)]
+        [CustomComboInfo("Simple Proc Fishing","Add Acceleration and Swiftcast to Simple Red Mage for proc fishing when necessary.", RDM.JobID)]
```
Iaotle, line 1037: Needs a better description I think.

Iaotle (changes requested):
Overall good, but pls fix some of the minor stuffs

XIVSlothCombo/Combos/RDM.cs
augporto: I thought of the steps because we need those skills to be called in a correct order for the opener to work, I agree it goes a bit on the heavy automation side, without the steps we would just call the skills when they are up and that would mess up the opener order or we would need specific control variables to know when to call what in order, and then it would end up at the same place.

If that is too much I can remove it np.

XIVSlothCombo/Combos/RDM.cs
augporto: Yep , didnt want to change it too much from the original , the macro would be better, going to change it.

XIVSlothCombo/Combos/RDM.cs
augporto: I use it 3 times actually, did the var to just calculate it once per invocation, did that because I dont know the cost of that

XIVSlothCombo/Combos/RDM.cs
augporto: Thought the levels maps where by job files and not centralized, at least the ones I looked at.

XIVSlothCombo/CustomComboPreset.cs
augporto: I agree, updated it with a better description.

XIVSlothCombo/CustomComboPreset.cs
augporto: I agree, updated it with a better description.

XIVSlothCombo/Combos/RDM.cs
Iaotle: I didn't see the other two times with Ctrl f haha no worries

XIVSlothCombo/Combos/RDM.cs
Iaotle: Yeah I mean the AST changes. Maybe do it when you get around to it

Issue #266, opened 2022-02-16
Semi-auto Ninjutsu Mudra for Ninja (No Bunny Feature / No Bunny Alternate Feature)


NikkoToska:
I think it should be possible to semi-automate casting ninjutsu so that you can't fail, if it's possible to track the Ninjutsu action state.  So if you can detect that after one mudra is pressed that Ninjutsu has become Fuma, then this should work:
```
Initial State -> 1st Press -> 2nd Press -> 3rd Press
----------------------------------------------------
Ten           -> Jin       -> Chi       -> Ninjutsu
Chi           -> Ten       -> Jin       -> Ninjutsu
Jin           -> Chi       -> Ten       -> Ninjutsu
```
This way it's easier to activate ninjutsu and should be impossible to mess up.  You might choose the wrong one, but if you just tap one button, the ninjutsu will go off.

So, for example:
For Fuma, I can press any mudra once, and I can press the Ninjutsu button to activate.
For Hyoton, I tap the Ten button twice, then press Ninjutsu to activate.
For Doton, I tap the Ten button three times, at which point all mudra become Ninjutsu, so one more tap will activate.

I think it would be good if there was an alternate version as well in the case that some people find it easier or more convenient to press a different button to start.  So the alternate would be:
```
Initial State -> 1st Press -> 2nd Press -> 3rd Press
----------------------------------------------------
Ten           -> Chi       -> Jin       -> Ninjutsu
Chi           -> Jin       -> Ten       -> Ninjutsu
Jin           -> Ten       -> Chi       -> Ninjutsu
```
It's only semi-auto since you still have to activate the tier two ninjutsu with a different button, but as long as you only press the same button, the ninjutsu won't fail.

nativepapaya:
This is pretty neat!

Iaotle:
Not sure if this is possible with the current toolset.

Iaotle:
@augporto u think this is possible?

augporto:
I think that may be possible yes, since ninjutsus are now gcds (I think), ill take a look at it as soon as I can.

augporto:
I made a PR that should implement the 1st example @NikkoToska proposed, if demand appear for the second one ill make it.

It need some feed back for high lv ninjutsu , the ones at lv76, since my ninja is not that high lv , but it should account for it.

NikkoToska:
[...]

PR #272, opened 2022-02-23
NIN Simple Mudra 

augporto:
Greetings,

### What it does
- Implement a preset for casting ninjutsu without failing
    - semi-auto
        - Lv1 should still be called with ninjutsu.
        - Lv2 can be casted with ninjutsu or any of the other 2 mudras , they will change to ninjutsu to avoid miscast.
        - Lv3 should make all mudras change to the ninjutsu.

- should solve #266 
    -  1st example, will try to make the 2nd example later.

### Known issues
- During my tests I have not met any, but feedback will be necessary for the high lv special ninjutsu's , lv76+ , since my ninja is not that high xD

Fun fact
The mudras are oGCDs disguised as GCDs, and you cant track them at all after the initial state, you have to track the ninjutsu that come from using the "combo" mudras.

ps: I need to rebase my fork , the commit list is starting to get out of hand >.>

### DIFF
--- a/XIVSlothCombo/Combos/NIN.cs
+++ b/XIVSlothCombo/Combos/NIN.cs
@@ -17,9 +17,6 @@ namespace XIVSlothComboPlugin.Combos
             DeathBlossom = 2254,
             AeolianEdge = 2255,
             TrickAttack = 2258,
-            Ninjutsu = 2260,
-            Chi = 2261,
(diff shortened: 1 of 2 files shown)


Iaotle:
In future might want to try a version bump included in the PR ;)

Good job!

PR #275, opened 2022-02-26
NIN add easy Mudra for Raiton when Lvl < 45

Spazzze:
Hi! 
I love your new feature "Simple Mudras" for NIN, but I have low-level NIN and I found out that it's impossible to cast Raiton with this feature enabled, because I don't have Jin learned yet. Can we add such an option?

image


Iaotle:
welcome! thanks for the fix

Issue #284, opened 2022-03-05
[RDM] oGCD Feature Everywhere


hokeenthong:
First of all, I must thank you guys again for this addon. I cannot thank you guys enough and I am thanking you again. 

This oGCD feature is good. However, the oGCD (Contre and Fleche) should be used after the instant cast of Veraero III or Verthunder III. Currently the button will change to oGCD once it is available and could change in the middle of casting jolt II. Thus, it will cause the button to do Jolt II > Contre or Fletche > Veraero III. For it to be efficient, it should do Jolt II > Veraero III > Contre or Fletch. Could you add a logic to it where dualcast is available, or when casting jolt II / verstone / verfire, the following ability should not change to oGCD. 

Thanks.

augporto:
I added a fix for it to the BRD PR I made, should solve those cases, got to wait it to be merged.

hokeenthong:
Thank you. Look forward to see it in action. Thanks again!!!

hokeenthong:
Thanks @augporto !! Love the update.

PR #286, opened 2022-03-06
[BRD] Couple tweaks to Simple Bard and RagingJaws features;

augporto:
Hello

### Modified

   - Simple Bard
       - Most notable:
           -  Added an option in the config window to customize the time which the snapshot happens relative to the remaining time of raging strikes.
                - Had the need to change that when using different attack skill spds, the reasoning behind it was that i had to change it almost every time i changed my threshold to avoid clipping etc @_@
           - Tweak on the opener to make compatible with lv80+ instead of only lv 90.
           - Tweak on pooling feature of BLs inside AP, start pooling at 30s on the song, to squeeze a little more dps.

### Note

 The custom time config , can be extended to any jobs that need a custom time for anything. I looked out on the other jobs that did something similar, but all of them shared the same variable , so changing the config in one changed on the other potentially messing up with each other, needed something that could store different values for different jobs, and the specification detached from central configuration.

```
public static class Config
        {
            public const string
                RagingJawsRenewTime = "ragingJawsRenewTime";
        }
```

on any job class representing a unique name for the config and call the config with:

`Service.Configuration.GetCustomConfigValue(BRD.Config.RagingJawsRenewTime);`

The code for the UI can be found in the ConfigWindow.cs, it stores the config on disk together with the others.

It only accepts configurations that represent time, a float.

### DIFF
--- a/XIVSlothCombo/Combos/BRD.cs
+++ b/XIVSlothCombo/Combos/BRD.cs
@@ -555,7 +555,7 @@ namespace XIVSlothComboPlugin.Combos
 
                 var canWeave = CanWeave(actionID);
 
-                if (IsEnabled(CustomComboPreset.BardSimpleOpener) && level >= 80)
+                if (IsEnabled(CustomComboPreset.BardSimpleOpener) && level >= 70)
                 {
                     if (inCombat && lastComboMove == BRD.Stormbite && !inOpener)
                     {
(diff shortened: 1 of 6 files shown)


augporto:
Last minute addition 
    - RDM #284 
        - Should fix this issue.

augporto:
Added a little modification to simple bard opener to account for lv 70+ , making it usable on ultimate fights.

PR #290, opened 2022-03-13
BLM Level Sync Fixes

grimgal:
Correct BLM levels.
Update BLM to cast Thunder/Thunder3 only if enough MP.
Update BLM to cast Fire3 when in Umbral Ice and max MP pre-Fire4.
Update BLM to cast Blizzard when in Umbral Ice and pre-Blizzard4.

### DIFF
--- a/XIVSlothCombo/Combos/BLM.cs
+++ b/XIVSlothCombo/Combos/BLM.cs
@@ -58,9 +58,9 @@ namespace XIVSlothComboPlugin.Combos
         public static class Levels
         {
             public const byte
-                Fire3 = 34,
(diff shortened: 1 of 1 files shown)


nativepapaya:
Would it be possible to adjust the bliz3 cast for when the user has less than 1600 mp and does not have despair? As it currently stands, doing the rotation you usually end up with around 1400 or so mp, which is not enough to cast another Fire IV.

grimgal:
Sure, I'll test out the rotation on level 70/80 just to check those rotations look good as well.

Iaotle:
Good work!

PR #297, opened 2022-03-23
[RPR] Soul Feature Fixes + AoE Combo Fixes

k-kz:
`SoulSliceFeature` and `SoulScytheFeature` now function correctly, and no longer interrupt Enshrouded burst phase.
Full combos/rotations should now work with these features enabled.

Fixes to Whorl of Death feature - now prioritised over Enshrouded if >10s of burst left.
Should also fit into combos/rotations correctly alongside Soul features.

Further testing required for fine-tuning, but feature priority should be correct now, and combos should no longer break.


Iaotle:
Thanks! In future, you can bump the version number in the PR to make it a bit faster to push changes.

PR #300, opened 2022-03-24
Added two options to RPR Slice combo: a Stun (Leg Sweep) and a Heal (Bloodbath & Second Wind)

BaexTower:
Two more options for the Reaper Slice combo.  Bloodbath and Second Wind as panic heals based on HP % and added leg sweep to check for interruptible cast.


k-kz:
These features were already in the pipeline for a big RPR update we just pushed (#302) and are now live in-game.
Also for future ref, the order of the code matters and I *think* having those features added at the end (at least the stun) would be non-functional (?)

Anyways, implemented now. Enjoy!

PR #324, opened 2022-03-30
Custom CanWeave with default 0.7 , retaining functionality

augporto:
- Based on the work Ele did on their PR for AST rework!


k-kz:
Thanks for the initial work @ele-starshade !!

PR #339, opened 2022-04-01
Update WAR.cs

juushin0:
Optimal Surgest tempest change


Iaotle:
I don't remember why this was changed but I do believe there was a reason for it.

NikkoToska:
The new slider on the Storms Path Combo allows you to set it to however many seconds you would like from 0 to 30.

PR #363, opened 2022-04-04
Added RedMageJoltVerprocCombo

Tsusai:
Added RedMageJoltVerprocCombo. Jolt 1&2 become a Verproc based on mana gauge's balance and available proc.
Might need double checking with possible conflict combos


Tsusai:
After some thinking, I want to revise this a bit more. I'll resubmit later

PR #432, opened 2022-04-08
MCH : Raising protection for clipping during hypercharge for reassemble;

augporto:
### MCH
   - Raising protection parameter for reassemble clipping;
   - Adding the ability to "late weave" the 1st wildfire;


augporto:
Not ready, made a mistake :O

PR #517, opened 2022-04-18
[MCH] Level check and better handling of oGCDs

StevieCoops:
Added a level check for Heat Blast on main combo and better handling of Gauss/Ricochet logic


k-kz:
@StevieCoops I merged this initially, but there were a couple missing parentheses etc that broke the code (which I fixed), but then @augporto requested to take a look over the commits tomorrow - please reopen this PR if it'll still let you and we'll get to it soon. Thanks!

augporto:
@k-kz from what i see this should not break anything , so it is/was ok to merge. I dunno what made the build bork that time.

k-kz:
> @k-kz from what i see this should not break anything , so it is/was ok to merge. I dunno what made the build bork that time.

Well, it was some missing parentheses that broke the code, which I fixed - I only put it on a brief hold so you could give it a look-over yourself as our MCH guy. ✌️

PR #535, opened 2022-04-20
Adjust SMN feature description, and tweaks.

Genesis-Nova:
Sets SMN gemshine feature to be a suboption of single target egi.


Genesis-Nova:
Smn single target egi options looks like this after commit.
image

PR #536, opened 2022-04-20
PLD: Sliders and Atonement fix

Nyuuuwu:
PLD:
Sliders for Intervene and Atonement are back
Atonement works as intended again


k-kz:
Addressed in #539

PR #730, opened 2022-05-19
SMN ST/AOE parity changes

Genesis-Nova:
- Moved `Demi Feature` above `Egi Feature` for single target as demis should have priority over egis (was this way already for AoE)
- Removed redundant level checks (checked for `Swiftcast` level, and `Ruby Ruin` level. `Swiftcast` is higher level)
- Applied `Egi Order`, `Egi Swiftcast`, and `Burst Phase` to AoE as well for parity.
- Changed order of features to account for `Egi Order`, `Egi Swiftcast`, and `Burst Phase` not being inside single target.
- Renames features that aren't ST only anymore to attempt to clarify what they do at first glance.
- Adds options for `Searing Light`, `Swiftcast`, and damage oGCDs to only use them on single target.

image

### DIFF
--- a/XIVSlothCombo/Combos/SMN.cs
+++ b/XIVSlothCombo/Combos/SMN.cs
@@ -269,9 +269,9 @@ namespace XIVSlothComboPlugin.Combos
                         {
                             if (gauge.HasAetherflowStacks)
                             {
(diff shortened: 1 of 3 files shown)


Genesis-Nova:
image

Genesis-Nova:
image

PR #739, opened 2022-05-21
Revert of DrawRadioButton commit

Tsusai:
After testing, ImGUI Radio buttons default (regardless of GUI showing no checkboxes or both) to the lowest number.  However, if you want to use a true default (checked in gui on show), start with Zero, not a One. Test results here: https://media.discordapp.net/attachments/977323921277849600/977358185272864778/unknown.png (0&1 and 1&2 on the first example, not 0&2)


Genesis-Nova:
Also going to point out that when the 'multiple options checked by default' (fresh config, no manual changes)
it does actually behave as if 'multiple' of the options that should be exclusive to one another enabled.

PR #745, opened 2022-05-22
Card Draw Targetting speedup

Tsusai:
AST: Moved Job names into CustomCombo as List groups (JobNames.Melee for example). In case it's needed in the future, any other feature can check the names.

AstrologianCardsOnDrawFeaturelikewhat SetTarget: Checking job names via new lists

Before info:
EN melee name search: 22 ticks average 
Same, but EN is after CN and JP: 50 ticks average

After: 
EN after CN & JP: 2 ticks average (vs 50)


k-kz:
Great work on this one. Thank you!

PR #757, opened 2022-05-27
Smn tweaks

Genesis-Nova:
Underlying logic tweaks, fixed bug with egi summon order


k-kz:
Looks good 👍
Leaving this as a note to self that I've reviewed already, will merge when I can

PR #778, opened 2022-06-04
Bugfixes Galore + Backend Tweaks

### LINKED ISSUE #683: [WHM] Thin Air Raise Feature + Global
Using the WHM Thin Air Raise Feature with the Global Healer: Raise Feature makes you cast cast Raise before Thin Air when spamming the button.

It works correctly if you use it with WHM Alternative Raise Feature.

### LINKED ISSUE #695: [PVP] [DRG] Dragoon LB + Emergency heals feature
While you have emergency heal feature and use Dragoon LB and your hp drops so the feature replaces all your skills, as well as LB skill to drop down from air, getting you stuck in LB untill it auto ends.

### LINKED ISSUE #735: [MCH] Ricochet and Gauss round overcap protection does not work when under lv 74
It will not leave 1 charge left of either.

### LINKED ISSUE #765: [NIN] Simple Mudras bug
The mudra path of simple mudras has an issue in my 70lv ninja
The Mudra Path set 1 of Ten Mudras which is supposed to be: Fuma Shuriken > Raiton/ Hyosho Ranryu, Suiton 
But it delivers this sequence: Fuma Shuriken > Hyoton > Hyoton

Actually, It was no problem yesterday before I wrongly clicked the update button

### LINKED ISSUE #767: "Hide to Mug" on NInja adds Trick Attack to the button
[SlothDebug.txt](https://github.com/Nik-Potokar/XIVSlothCombo/files/8795265/SlothDebug.txt)

### LINKED ISSUE #775: [NIN] addon randomly stops using mudras in aoe combos
as title says. The addon will randomly stop using mudras during the aoe combo. Meaning doton and katon wont be used.
Using a mudra manually will sometimes fix the issue.

Taurenkey:
[Bug Fixes]
- Fixes #775 Should be fixed due to now undoing the previous changes.
- Fixes #767 Preset description adjusted and removed the `Suiton` check as 6.1 changed functionality of `Trick Attack` making this less useful for this feature.
- Fixes #765 All routes now correctly use the Mudras described.
- Fixes #735 Changed the check from cooldown remaining to checking for more than 1 charge.
- Fixes #695 Emergency features now won't kick in whilst under the effect of `Sky High` (the DRG LB buff they get whilst in the air)
- Fixes #683 Global feature tweaked to be more responsive by removing the check for the `Swiftcast` buff before returning the appropriate rez spell (or WHM's `Thin Air` if that feature is enabled) but still checks if `Swiftcast` is off cooldown to return that.

[Framework]
- Adds a way to check for `Pacification`, `Silence` and `Amnesia` (the 3 debilitating statuses in PoTD & HoH). `HasPacification()`, `HasSilence()` and `HasAmnesia()`.
- `GetResourceCost` has been added to check the value an action _may_ cost. Does not take into account free casts, but does take into account increased MP costs such as BLM's `Astral Fire`.
- Adds a class check to prevent players trying to use features on a class when they should be on the job instead.

### DIFF
--- a/XIVSlothCombo/ActionWatching.cs
+++ b/XIVSlothCombo/ActionWatching.cs
@@ -9,11 +9,11 @@ namespace XIVSlothComboPlugin
 {
     public static class ActionWatching
     {
-        private static Dictionary<uint, Lumina.Excel.GeneratedSheets.Action>? ActionSheet = Service.DataManager?.GetExcelSheet<Lumina.Excel.GeneratedSheets.Action>()?
+        public static Dictionary<uint, Lumina.Excel.GeneratedSheets.Action>? ActionSheet = Service.DataManager?.GetExcelSheet<Lumina.Excel.GeneratedSheets.Action>()?
             .Where(i => i.RowId is not 7)
             .ToDictionary(i => i.RowId, i => i);
 
-        private static Dictionary<uint, Lumina.Excel.GeneratedSheets.Status>? StatusSheet = Service.DataManager?.GetExcelSheet<Lumina.Excel.GeneratedSheets.Status>()?
+        public static Dictionary<uint, Lumina.Excel.GeneratedSheets.Status>? StatusSheet = Service.DataManager?.GetExcelSheet<Lumina.Excel.GeneratedSheets.Status>()?
             .ToDictionary(i => i.RowId, i => i);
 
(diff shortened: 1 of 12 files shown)


k-kz:
Fixed conflicts with a `NIN` skill in `CustomComboPreset` and the `WHM` `Thin Air Raise` fix.

PR #782, opened 2022-06-06
[NIN] Simple Fixes

Tsusai:
NinAeolianAssassinateFeature: Cleanup
NinAeolianMugFeature: Simplified
NinjaArmorCrushCombo: Fixed hard level calls
NinHuraijinArmorCrush: Simplified If
NinjaHideMugFeature: InCombat API call updated to InCombat function
NinjaKassatsuChiJinFeature: Fixed hard level call

### DIFF
--- a/XIVSlothCombo/Combos/NIN.cs
+++ b/XIVSlothCombo/Combos/NIN.cs
@@ -123,7 +123,7 @@ namespace XIVSlothComboPlugin.Combos
 
             protected override uint Invoke(uint actionID, uint lastComboMove, float comboTime, byte level)
             {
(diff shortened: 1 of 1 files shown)


Taurenkey (changes requested):
Please do not change the values assigned to configs, you can keep the variable name changes but changing the values themselves resets user configs which we want to avoid doing where necessary.

k-kz:
Oops. I saw strings and just.... Yeah. Will sort that in a bit hahah.

Tsusai:
Removing this one as I think I borked this branch.

PR #783, opened 2022-06-06
[PLD] Simple Tweaks

Tsusai:
PaladinRoyalAuthorityCombo: Updated to use InCombat() function instead of having API code. interveneCharges var moved nearer it's use
PaladinProminenceCombo: API -> InCombat() function


Tsusai:
Ignore the wtfery it's 23:45 and i'm not thinking right

Tsusai:
Will redo later

PR #791, opened 2022-06-09
[WHM] Lily fix.

cupcake121:
Puts Misery before Medica 2.

Misery -> Blood Lily (if proced) -> Medica 2 -> Medica 1


cupcake121:
Puts Misery before Medica 2.

Misery -> Blood Lily (if proced) -> Medica 2 -> Medica 1

PR #807, opened 2022-06-14
Number to Class Constant Level Check Fixes

Tsusai:
Updated various level checks to use class levels constants instead of typed number

- [x] BLM
- [x] MCH
- [x] MNK
- [x] ~NIN~

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -130,6 +130,11 @@ namespace XIVSlothCombo.Combos
             [ParentCombo(ALL_Melee_Menu)]
             [CustomComboInfo("Melee DPS: Double Feint Protection", "Prevents the use of Feint when target already has the effect by replacing it with Fire.", ADV.JobID)]
             ALL_Melee_Feint = 100030,
+
+            [ParentCombo(ALL_Melee_Menu)]
+            [CustomComboInfo("Melee DPS: True North Protection", "Prevents the use of True North when its buff is already active by replacing it with Fire.", ADV.JobID)]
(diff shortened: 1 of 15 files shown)


k-kz:
## BLM
All good, just `90` --> `Paradox`.

## MCH
Putting this on hold until @augporto can review - you may notice that on lines 142 - 146, the levels `76` and `75` are used - <br>I'd assume this was for good reason so will wait for confirmation.
image

## MNK
I'm also guessing @grimgal has done MNK's lines 624 - 638 in a similar manner, and it seems like you've probably killed some logic there for the sake of tidiness (at a glance, at least, I don't know how MNK works).
image

## NIN
Also "rejecting" this on the basis that @Taurenkey is currently doing a NIN rework. <br>Although the code itself here does look good 👍🏼

Tsusai:
NIN: I'll revert shortly

MNK: Riddle Of Fire is 68. Ln 629's inbetween level check is sort of redundant, because even at the end, it'll fall back to RoF.  The only condition in which Brotherhood will show is Ln 633, which is what the new code boils down to. If I'm missing something in following the logic (which has happened) I hope @grimgal lets me know.

MCH: Yup. Forgot to make it < instead of <= on the hotshot line (Hotshot becomes Air Anchor at 76).  Although I think it can be simplified further using OriginalHooks but would need testing (not the scope of these level fixes atm). I'll get that tweaked mathematical issue fixed. AFAIK It's main combo, where I believe @augporto was working on the Simple?

augporto:
@k-kz  as far as i can see, that is right , just that `<=` should be only `<` , but @Tsusai already caught up to it. So it should be fine after he corrects that o/

PR #830, opened 2022-06-26
[Framework] TargetExtensions

Tsusai:
Taurenkey "Now make them extension methods"
and lo, I was bored enough to do so
90% of Target related functions from Target.cs have been converted to extension methods
SGE has the proof of concept + Taurenkey's LevelChecked() extension
Party SetAsTarget not added due to protected status my brain has yet to comprehend.


Tsusai:
Mommy and Daddy are fighting

PR #834, opened 2022-06-28
[ALL] Bugfix for Global Raise & SCH/SMN

Tsusai:
ALL_Healer_Raise was affecting Summoner & ALL_Caster_Raise was affecting Scholar due to both classes having the same Resurrection ID. This has been fixed by this PR, _**BUT**_ needs a review. LocalPlayer.ClassJob.Id works as something to check but if there is a better item to check against SCH/SMN.JobID, please let me know


k-kz:
@Taurenkey I'm merging this in, @Tsusai has tested and it looks good to me - but we were both wondering if you had any ideas re: a 'smarter' way about this

PR #836, opened 2022-06-29
Fixed MNK Demolish/Twin Snakes slider not working correctly.

Genesis-Nova:
Changed `GetCustomIntValue` to `GetCustomFloatValue` since the value is a float.


Genesis-Nova:
Because that kind of micro `endgame`-only optimization is not something we're looking to focus on with our plugin.

PR #850, opened 2022-07-02
[BRD] `LevelChecked()` + Cleanup

k-kz:
## Changes
Removed whitespace
Migrated level checks to `LevelChecked()`
Comment spacing
Removed level constants
Removed redundant parentheses
Added regions for var sets

@augporto could you take a glance at this one when you get a min? Ty

## To-Do
- [x] Migrate to `LevelChecked()`
- [x] File cleanup (whitespace, comments, etc)
- [ ] Test


augporto (approved):
From what i can see all is ok , it should work 😄

PR #868, opened 2022-07-08
[SGE] Bugfix / Light Cleanup

Tsusai:
SGE
* Actions made internal vs public
* Gauge made private, should point to function instead of being function result (Bugfix?) 
* SGE_ST_Dosis
  * Toxikon: Clarified/Renamed Toxikon bool per Discord discussion w/ k
  * Bugfix: Eukrasia buff check moved higher to encompass DoT & Toxikon code. Toxikon was interrupting Eukrasia/Eukrasia Dosis two button combo.
* SGE_AoE_Heal 
  * Bugfix. Didn't do OriginalHook on Physis correctly with ActionReady
  * Eukrasia buff check moved to the top to stop replacement actions from breaking the Eukrasia->Eukrasia Prognosis combo
* SGE_ST_Heal:
  *  Haima was looking for current target still. Fixed. Ingame configuration will now show the options (Haima, Zoe, etc) in the same order as the code
  * Eukrasia buff check moved to the top to stop replacement actions from breaking the Eukrasia->Eukrasia Diagnosis combo

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -2173,11 +2173,11 @@ namespace XIVSlothCombo.Combos
         #region Diagnosis Simple Single Target Heal
         [ReplaceSkill(SGE.Diagnosis)]
         [ConflictingCombos(SGE_Rhizo, SGE_DruoTauro)]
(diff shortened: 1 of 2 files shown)


Tsusai:
Todo:

- [ ] Check if Toxikon interrupts Dosis DoT "combo" while moving fix

PR #889, opened 2022-07-11
[SGE] Bugfixes

Tsusai:
FRMWK
* Added HasFriendlyTarget. 

SGE
* Fixed Gauge reference. No functional change but I meant to do a method forward and not a class variable 
* SGE_ST_Dosis_EDosis: Resovlved edge case error where EukrasiaDosis is still available in the half second Eukrasia buff is removed
* SGE_ST_Heal: Implemented HasFriendlyTarget to allow SGE_ST_Heal to work again with AI party members/chocobos


Tsusai:
Solution for ST Heals identified, marking as draft until tested

PR #895, opened 2022-07-15
[DRG] Cleanup

damolitionn:
Cleaned up simple DRG. Renamed to Advanced.
Removed all the other combo options.
Dives option now on CD, double weave, or single weave.
Added Stardiver Feature: Stardiver turns into Nastrond during Life of the Dragon and Geirskogul outside of it.
Added Lance Charge to Battle Litany Feature, and Dragon Sight sub option.
Added Melee feature for dives.
Redid DRG config. Remind users to check settings.


k-kz:
> Redid DRG config. Remind users to check settings.

@damolitionn did you want us to put out a login message type warning, or actually reset some of the users' configs for this?

damolitionn:
Either option is fine I think as long as users are made aware of the changes to the config.

PR #909, opened 2022-07-23
GNB ST/AoE Combo Starter

kawaii:
To bring the job in parity with others as discussed in the Discord; changes the ST and AOE combos to use their first skill in the chain as the combo starter, instead of the last.


k-kz (changes requested):
Thanks for chucking this together - could you have a sweep through the other GNB features/options and double-check that there's nothing else referencing the `Solid Barrel Combo` in descriptions that needs conversion?

k-kz:
Closing due to these changes being included in #1078

PR #917, opened 2022-07-29
[DNC] Option to add Flourish to Fan Dance 1/2

jundong321:
DNC
- Option to add Flourish to Fan Dance 1/2
- Reprioritize Fan Dance 3 in ST/AoE MultiButton


k-kz:
Won't be merging this:
- Altered existing feature IDs, which messes with existing userconfig
- Loads of redundant level checks added to buff checks (in which you cannot have the buff without the appropriate level)
- `FD1`/`FD2` --> `Flourish` is something I can add as a part of the upcoming DNC Advanced/Simple rework, but as a multi-checkbox, rather than two options
- Esprit overcap additions will also be addressed in a 'main combo context' in the rework

PR #919, opened 2022-07-31
[SCH] Scholar feature improvement

jundong321:
New Features:
- SCH_Lustrate: Change `Lustrate` into `Excogitation` when `Excogitation` is ready.
- SCH_Recitation: Change `Recitation` into one of `Adloquium`, `Succor`, `Indomitability`, and `Excogitation`.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -2647,8 +2647,8 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Fey Blessing to Seraph's Consolation Feature", "Change Fey Blessing into Consolation when Seraph is out.", SCH.JobID, 210, "", "")]
         SCH_Consolation = 16210,
 
-        [ReplaceSkill(SCH.Excogitation)]
-        [CustomComboInfo("Excogitation to Lustrate", "Change Excogitation into Lustrate when Excogitation is in cooldown.", SCH.JobID, 220, "", "")]
+        [ReplaceSkill(SCH.Lustrate)]
+        [CustomComboInfo("Lustrate to Excogitation", "Change Lustrate into Excogitation when Excogitation is ready.", SCH.JobID, 220, "", "")]
         SCH_Lustrate = 16220,
 
         [ReplaceSkill(SCH.Recitation)]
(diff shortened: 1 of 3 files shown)


k-kz:
Will be taking a proper look at this with @Tartarga once #908, #914 and #926 are merged - it looks good for the most part but we'll need to do some conflict avoidance after his stuff goes through, I'm pretty sure.

Tartarga:
@jundong321 Lustrate looks good to me, but I have a question on Recitation. Are the options for Recitation supposed to be a "Choose only one of the four options" sort of thing? (Comparing your code with your description)

k-kz:
> Lustrate looks good to me, but I have a question on Recitation. Are the options for Recitation supposed to be a "Choose only one of the four options" sort of thing? (Comparing your code with your description)

To add to this, it seems like it might be a better approach to create a multi-checkbox which allows `Recitation` to be placed over any combination of the four skills - that way, you can add it to the skills you'd like and use it in a more flexible manner when in combat by double-tapping the relevant action.

(I should probably note that I don't play SCH and might be misunderstanding the intended use, but I'm under the impression that `Recitation` is to be used before the other skills in order to buff them in some way)

jundong321:
> > Lustrate looks good to me, but I have a question on Recitation. Are the options for Recitation supposed to be a "Choose only one of the four options" sort of thing? (Comparing your code with your description)
> 
> To add to this, it seems like it might be a better approach to create a multi-checkbox which allows `Recitation` to be placed over any combination of the four skills - that way, you can add it to the skills you'd like and use it in a more flexible manner when in combat by double-tapping the relevant action.
> 
> (I should probably note that I don't play SCH and might be misunderstanding the intended use, but I'm under the impression that `Recitation` is to be used before the other skills in order to buff them in some way)

Thanks for the suggestion! What you described is one approach, and I think it's already there. The other approach is to have recitation changed to one of the four skills, so that you can double tap the recitation button to use the combo. It will allow only one of the combos though, so the flexibility is limited.
The reason I choose this approach is that I try to avoid covering the healing skills with other skills, so that the button press is a more defined behavior for these skills. Like sometimes I press two times to heal and sometimes I press one time to heal.
But I totally agree that this is a choose one of the four feature, and would be nice to have an exclusive select. I will add feature conflicts to do that.

k-kz:
Please feel free to send a message on discord if you'd like to talk with us about it in greater depth :)

jundong321:
@Tartarga the suggested code change is in, tested working as expected.

Tartarga:
Just tried it out, everything plays great, but the only issue I had was Excog and Succor values being mixed (top is userconfig.cs, bottom is SCH.cs)
image

PR #964, opened 2022-09-06
[BLM] Simple AOE Feature

monkeymc:
From discord https://discord.com/channels/957680331056615506/957948813585317938/1015639550078505113

I think because Fire II need MP 3,000

image

### DIFF
--- a/XIVSlothCombo/XIVSlothCombo.csproj
+++ b/XIVSlothCombo/XIVSlothCombo.csproj
@@ -66,7 +66,7 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="DalamudPackager" Version="2.1.8" />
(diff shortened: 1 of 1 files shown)


monkeymc:
I have new feature from request in #BLM ref: https://discord.com/channels/957680331056615506/957948813585317938/1017885280184844299

Daemeous (commented):
Perhaps changing "Use Thunder" to "Use Thunder II/IV", "Add Thunder II/IV when the debuff isn't present or expiring." Would be a good idea, perhaps giving it parity with the Thunder usage in the Simple BLM feature's Single Target, splitting it into Thundercloud and Thunder II/IV separately.

As for a description for the who slider bars:
"Will recast Thunder II/IV if the remaining duration is lower than this many seconds."
"Will not cast Thunder II/IV on targets below this HP % threshold."

PR #965, opened 2022-09-06
Post Raid Adjustments

damolitionn:
GNB (Thoroughly tested through the tier):
- Bloodfest cartridge management removed.
- Adjusted opener for 1NM23GF Opener.
- Adjusted timing for 1GCD GF delay at 2 min window.
- Removed GFNM logic from 6.1.
- Added !IsMoving check to melee Rough Divide option.
- Adjusted 30 sec Danger Zones when forced to delay the skill.
- Adjusted Bloodfest burst windows with 6.2 changes.

DRK:
- Reduced Blood overcap management during Blood Weapon from 70 to 60.

DRG:
- Bugfix where Piercing Talon would deactivate the opener if it was the first ability used in combat.
- Fixed Piercing Talon Option on AOE.

RPR:
- Added Opener option that allows for Early Enshroud and Early Gluttony Openers.
- Adjusted Soul management so that Soul is pooled for Gluttony so as to stop Gluttony drift.
- Adjusted Enshroud windows for natural odd/even Enshrouds.

SAM:
- Bugfix where Enpi would deactivate the opener if it was the first ability used in combat.
- Various adjustments the opener.
- Tightened the windows that begin and end the filler windows.
- Fixed Meikyo Burst option.

WAR:
- Added !IsMoving to melee Primal Rend and Onslaught options.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1972,7 +1972,7 @@ namespace XIVSlothCombo.Combos
         RPR_ST_SliceCombo_Enshroud = 12008,
 
         [ParentCombo(RPR_ST_SliceCombo_Enshroud)]
-        [CustomComboInfo("Enshroud Burst (Double Enshroud) Option", "Uses Enshroud at 50 Shroud during Arcane Circle (mimics the 2-minute Double Enshroud window), but will pool Shroud outside of burst windows.\nBelow level 88, will use Enshroud at 50 gauge.", RPR.JobID, 0, "", "")]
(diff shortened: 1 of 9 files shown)


GibXIV:
RPR "Added Opener option that allows for Early Enshroud and Early Gluttony Openers."

With opener option (either early gluttony or early enshroud), but without enabling Enshroud option and/or Arcane Circle option the opener will just look weird, pooling soulgauge and shroud but not spending properly.

Add information to CustomComboInfo to inform user that opener needs these options, and/or hardcode that when opener is enabled that enshroud and circle are forced on (similar to BLM opener forcing ley lines)

k-kz:
> Add information to CustomComboInfo to inform user that opener needs these options, and/or hardcode that when opener is enabled that enshroud and circle are forced on (similar to BLM opener forcing ley lines)

The latter suggestion here sounds like the best move, if those other options are essential for correct opener function

damolitionn:
Hard coded the RPR Opener now

PR #966, opened 2022-09-07
[SCH] Add SCH_Ruin2 that puts Bio on Buin2

jundong321:
[SCH]

New feature:
- SCH_Ruin2: Replaces Ruin II with Bio I/II for DoT Uptime


k-kz (changes requested):
We'll need to do a revert on the Feature ID changes in this PR to avoid changing up what users have toggled without them being aware, and at a glance this doesn't seem to warrant us forcing a reset.

jundong321:
@k-kz good point, I've reverted all feature id changes.

PR #972, opened 2022-09-07
BRD/DRG/MCH/MNK/NIN/SAM Self-Healing all updated.

Daemeous:
Updated and tested all jobs (except BRD which I don't have yet), lowered Nin priority quite a bit to avoid CD drift.

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRG.cs
+++ b/XIVSlothCombo/Combos/PvE/DRG.cs
@@ -69,7 +69,11 @@ namespace XIVSlothCombo.Combos.PvE
             public const string
                 DRG_ST_DiveOptions = "DRG_ST_DiveOptions",
                 DRG_AOE_DiveOptions = "DRG_AOE_DiveOptions",
-                DRG_OpenerOptions = "DRG_OpenerOptions";
+                DRG_OpenerOptions = "DRG_OpenerOptions",
+                DRG_STSecondWindThreshold = "DRG_STSecondWindThreshold",
+                DRG_STBloodbathThreshold = "DRG_STBloodbathThreshold",
+                DRG_AoESecondWindThreshold = "DRG_AoESecondWindThreshold",
+                DRG_AoEBloodbathThreshold = "DRG_AoEBloodbathThreshold";
         }
 
         internal class DRG_JumpFeature : CustomCombo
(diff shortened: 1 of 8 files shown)


XIVSlothCombo/Combos/PvE/DRG.cs
```diff
@@ -177,6 +181,15 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
 
                                 if (WasLastWeaponskill(HeavensThrust) && GetRemainingCharges(SpineshatterDive) > 0 && !WasLastAction(SpineshatterDive) && openerOptions is 0 or 1 or 2)
                                     return SpineshatterDive;
+
+                                // healing - please move if not appropriate priority
```
damolitionn, line 185: Why are these in the opener

XIVSlothCombo/Combos/PvE/SAM.cs
```diff
@@ -219,6 +223,15 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
 
                                 if (gauge.Kenki >= 25 && IsOnCooldown(Shoha))
                                     return Shinten;
+
+                                // healing - please move if not appropriate this high priority
```
damolitionn, line 227: Same with these

XIVSlothCombo/Combos/PvE/SAM.cs
```diff
@@ -482,6 +495,15 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
 
                                     if (IsEnabled(CustomComboPreset.SAM_ST_GekkoCombo_CDs_Shoha) && Shoha.LevelChecked() && gauge.MeditationStacks == 3)
                                         return Shoha;
+
+                                    // healing - please move if not appropriate this high priority
```
damolitionn, line 499: These are nested under `SAM_ST_GekkoCombo_CDs` which means your customcombopreset will need to be adjusted accordingly

Daemeous: Do you think that's the best way to do it? I was fretting over doing that, which would make it unique among all the implementations or having it's own separate canWeave location for it. But I must confess, the Sam code is a bit too big brain for me to figure out quite where that would be

damolitionn: It's either that or move it somewhere else. Users will need to have `SAM_ST_GekkoCombo_CDs` on as well then and if it's not nested then there's no indication that it's required to be on for your healing feature to work.

PR #975, opened 2022-09-08
Split up UserConfig file. UserConfigItems are now in the Combo Folder

Tartarga:
Git magic to keep history on both files


Tartarga:
Will revisit after .3 release

Issue #981, opened 2022-09-09
[MNK]  Meditation when out of range or without a target


ArcanaXVIII:
Hi everyone,

Just a simple suggestion for MNK.

Since it's the only melee job without a ranged attack, in lieu of a range uptime feature on main combo like the other jobs, it'd be nice if Meditation came up when out of range or without a target instead. 

As it stands we have "Meditation on main combo" (#309), but it only comes up out of combat for that instant refill. Getting back some meditation stacks while out of range/during untargetable boss phases would be a gain.  Still more useful than Anatman. :D

Thanks!

GibXIV:
Thank you for your idea @ArcanaXVIII i took a whack at it and my pullreques https://github.com/Nik-Potokar/XIVSlothCombo/pull/982 has been approved and merged, will show up on a next release ;)

ArcanaXVIII:
> Thank you for your idea @ArcanaXVIII i took a whack at it and my pullreques #982 has been approved and merged, will show up on a next release ;)

Yeah I saw the Pull Request go through, thanks for doing it!

PR #982, opened 2022-09-09
Monk add Meditation builder out of melee range

GibXIV:
#981 inspired me to make my first PR.

Added option to add Meditation chakra builder to Simple Single Target rotation.
image

Will only use it out of opener/builder, and out of melee range.
Did not add to AoE due to different AoE melee ranges.


Daemeous (changes requested):
Returning just Meditation ID causes issues in low level synced content, returning the Original Hook fixes this, so:
return Meditation; -> return OriginalHook(Meditation);

Genesis-Nova:
Well Meditation is lvl15, its building chakra not spending it.
I think you might be thinking of Steel Peak > Forbidden Chakra (54)

Issue #870, opened 2022-07-08
[Healers] Add DoT maintenance refresh sliders


happilysleeping:
i love the reaper "shadow of death" option where you can set X seconds remaining before refreshing death's design, and would love to have that on my sage/scholar!

Daemeous:
resolves #870

PR #1016, opened 2022-10-03
[Healers] Added DoT Thresholds

Daemeous:
Aki:
Added DoT refresh thresholds.

Daemeous:
Modified DoT refresh thresholds to float with 0.1s increments.


Daemeous:
resolves #870

Issue #1010, opened 2022-09-27
[GNB] Burst Strike into Double Down feature improvement


bronzo44:
Hello,
let me start with saying that I love this feature, it's the only thing I use Sloth for and I find it amazing, making me save a very needed keybind and at the same time making the BS/DD key feel so... Nice and powerful!

That said, there is one pretty big issue with this: using this feature forces you to drift DD by 1 GCD every minute: this happens because DD's cooldown comes off together with the GCD timer, hence the key is a Burst Strike all the way until the GCD ends, and only after that it becomes DD. So, if you spam the key, you'll fire off a Burst Strike, even if in that GCD DD is already up (but Sloth couldn't change the key in time).

Currently you can either clip the GCD to wait for Sloth's substitution (which will likely lose you the 9th NM gcd) or have a separate keybind for DD to use just occasionally, since you can afford with 2.42 or even 2.45 sets to drift almost all DDs and then reverse drift it with a couple off-CD usages.

Now, both solutions kill the purpose of Sloth. Moreover, the 2.50 set that's been recently proposed and that looks like it could be actually strong in a few fights this tier totally negates the possibility of drifting DD, since the reverse drift cannot happen at 2.50 - which means, if you drift it, you drift it for good and eventually it'll slip out of NM.

**What I think could be a reasonable solution** is to change how the BS-DD key substitution works, making it so the swap happens 1 or 2 seconds before DD comes off CD. This would allow players to have DD on the key before the GCD finishes rolling, and fire it off CD.

This is possible because there is no scenario in which, during NM, you want to fire off a stray BS when DD is coming off CD.

Sorry for the textwall, I wanted to be as precise as possible with this request. Thank you for your attention!

k-kz:
This may have been addressed in #965 (which is pending in the `v3.0.17.3` update) - it's out of my area of expertise so I'm not 100% sure.

@damolitionn

k-kz:
Addressed in #965

Issue #922, opened 2022-08-04
[RPR] Regress Delay


happilysleeping:
Another combo fork I used to use had a feature that could delay the use of Regress for a few (configurable) seconds, which saved me from fucking up a few times lol. I'd love to have that option in sloth.

tysm for ur work either way!!!

k-kz:
Addressed in #1017

PR #1017, opened 2022-10-03
Cleaning up some open issues.

### LINKED ISSUE #1001: [RPR] Gluttony overrides active Gibbet and Gallows
Hi, like the title said I came across a bug where Gluttony would override an active Gibbet and Gallows attack and thus losing it in the process. This wasn't the case in earlier versions but I'm afraid I can not pinpoint the time it changed although it has been a couple of weeks at least.

This happens if you don't use everything on one button button instead use an extra button only for Gluttony/Blood Stalk/Gibbet & Gallows (and the Enshroud combo but that's not relevant).

In short: Have a separate button for Blood Stalk/Grim Swathe, activate the option to have it be overriden by Gluttony, also activate the option to have Gibbet & Gallows on the same button -> Active Gibbet & Gallows attack gets overriden by Gluttony and will be lost in the process.

Picture of relevant config attached:
bug

### LINKED ISSUE #1002: [DRK] Delayed Bloodspiller Feature not in line with Balance Opener
Currently, the feature will delay Bloodspiller by 2 GCDs on even minute windows, which is contrary to the 3 GCDs in the general purpose opener, often missing buffs on the first cast.

### LINKED ISSUE #1005: [RPR] #422 Arcane Circle #431 scythe & #413 slice combo (bug)
Hello i have noticed that since 6.2 that #422 on slice combo is not activating at the first and sometime second application of death's design. Even going though an attire enshroud without it. Before 6.2 was always activating after death's design on start of a pull before Gluttony and so on.

As For Scythe & slice combo' It sometimes when Enshrouded and a mob dies then you switch mobs reapplies Death's design if not already applied but try's to go back to normal slice and scythe combo instead of void and cross or lemure's. 

I do have controller setting to auto target next available mob. If that info helps cause if I manually switch mobs it go's back to void and cross or lemure's

damolitionn:
Framework:
- Changed Horizontal Radio Buttons to radios for parity.

GNB:
- Moved opening NM to 3rd GCD because I'm not explaining my reasoning to you casuals.
- Adjusted timing for ```Double Down on Burst Strike Feature```, requested in #1010.

DRK: 
- Adjusted Bloodspiller dumps in ST, fixes #1002.

RPR:
- Added a 1 second delay to Regress, requested in #922.
- Added SoulReaver check to ```Gluttony on Blood Stalk/Grim Swathe Feature```, fixes #1001.
- Removed DD check on ST and AOE features, fixes #1005.

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRK.cs
+++ b/XIVSlothCombo/Combos/PvE/DRK.cs
@@ -144,7 +144,7 @@ namespace XIVSlothCombo.Combos.PvE
 
                             //Delayed Delirium
                             if (IsEnabled(CustomComboPreset.DRK_DelayedBloodspiller) && GetBuffStacks(Buffs.Delirium) > 0 &&
-                                (GetBuffStacks(Buffs.BloodWeapon) is 0 or 1 or 2))
+                                ((GetBuffStacks(Buffs.BloodWeapon) <= 2 && CombatEngageDuration().TotalSeconds < 30) || CombatEngageDuration().TotalSeconds >= 30))
                                 return Bloodspiller;
 
                             //Blood management before Delirium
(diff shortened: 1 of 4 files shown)


bronzo44:
"Adjusted timing for Double Down on Burst Strike Feature, requested in https://github.com/Nik-Potokar/XIVSlothCombo/issues/1010."

In the current released Sloth version it doesn't look like this was updated - the feature still works as previously, with BS changing into DD only when off CD completely

PR #1020, opened 2022-10-04
Bard and Samurai Combo Heal Fixes

Daemeous:
It was in totally the wrong place and AoE wasn't implemented correctly.


k-kz:
Looks good, at a glance on mobile I can see there's some stuff that can be changed to `ActionReady()`, though.

Daemeous:
They could, yes, but then they would be the only two actionReady() checks in BRD.cs.

k-kz:
> They could, yes, but then they would be the only two actionReady() checks in BRD.cs.

Changing towards using it is a good thing no matter when/where it happens. I don't see what the problem is

PR #1028, opened 2022-10-06
[Bugfix/Framework] ActionWatching LastAction fixes for non-English clients

Tartarga:
Checking what type an action is using row ID instead of English strings.

Any weird issues with LastAbility, LastSpell, LastWeaponskill not working might've been related if user's game was not set to English.


Tartarga:
Recommend a small MOTD with this along the lines of "If you've submitted a Github Issue and are not running the game in English, please retest those issues."

PR #1032, opened 2022-10-07
[NIN] Added `Dynamic True North Feature` to `Armour Crush`

Daemeous:
Given the nature of Ninja and this feature, it doesn't make sense to add it to anything else and it'll most frequently get overridden by other more important oGCDs.
This also changes the standard True North Armour Crush feature to delayed weave, which shouldn't change anything really because it'd weave early and just wait for Armour Crush GCD before anyway.
Separated out from the other branch in the likely event Taur, good boy that he is, doesn't (understandably) want my version defiling his glorious Ninja and implements it in a much better way.


k-kz:
closing due to some conflict merge shenanigans - PR reinstated in #1118

PR #1033, opened 2022-10-07
[SCH] Heal combo fixes + internal naming parity

Daemeous:
Daemeous:
Fix Scholar Healing and Fairy Combos.
Adds Consolation option to Fairy Combo for use under Seraph.
Changed Lucid Dreaming on Succour to be emergency in case of not enough MP.
Added sub-option to only use Aetherflow if Indomitability is ready,
Naming Parity established with Sage.
Clarified descriptions.

Tartarga:
Fixed Adloquium Bar.


XIVSlothCombo/Combos/PvE/SCH.cs
```diff
@@ -445,8 +445,7 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                     // Lucid Dreaming
                     if (IsEnabled(CustomComboPreset.SCH_Succor_Combo_Lucid) &&
                         ActionReady(All.LucidDreaming) &&
-                        LocalPlayer.CurrentMp <= Config.SCH_Succor_Combo_LucidOption &&
-                        CanSpellWeave(actionID))
+                        LocalPlayer.CurrentMp < 1000)
```
k-kz, line 448: Just noting for myself that this is an edge-case prevention on the off-chance the user doesn't have enough MP to cast `Succor` in the first place

PR #1040, opened 2022-10-08
[NIN] Uptime Functionality Readded

Daemeous:
Added back in Suiton Uptime functionality.
Reordered Uptime Features to reflect priority order.
Removed redundant !HasEffect(Buffs.Suiton) check as setupSuitonWindow already does that check.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1642,9 +1642,15 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Ninjitsu Option", "Adds Ninjitsu to Advanced Mode.", NIN.JobID)]
         NIN_ST_AdvancedMode_Ninjitsus = 10011,
 
+        [ConflictingCombos(NIN_ST_AdvancedMode_Ninjitsus_ChargeHoldForBurst)]
         [ParentCombo(NIN_ST_AdvancedMode_Ninjitsus)]
         [CustomComboInfo("Hold 1 Charge", "Prevent using both charges of Mudra.", NIN.JobID)]
         NIN_ST_AdvancedMode_Ninjitsus_ChargeHold = 10012,
+
+        [ConflictingCombos(NIN_ST_AdvancedMode_Ninjitsus_ChargeHold)]
+        [ParentCombo(NIN_ST_AdvancedMode_Ninjitsus)]
+        [CustomComboInfo("Hold 1 Charge For Burst/Uptime", "Prevent using both charges of Mudra except for burst or uptime features (if enabled).", NIN.JobID)]
+        NIN_ST_AdvancedMode_Ninjitsus_ChargeHoldForBurst = 10068,
 
         [ParentCombo(NIN_ST_AdvancedMode_Ninjitsus)]
         [CustomComboInfo("Use Fuma Shuriken", "Spends Mudra charges on Fuma Shuriken (only before Raiton is available).", NIN.JobID)]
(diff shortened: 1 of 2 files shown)


XIVSlothCombo/Combos/PvE/NIN.cs
```diff
@@ -371,8 +374,8 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                             if (IsEnabled(CustomComboPreset.NIN_ST_AdvancedMode_Ninjitsus_Suiton) &&
                                 setupSuitonWindow && 
                                 TrickAttack.LevelChecked() &&
-                                !HasEffect(Buffs.Suiton) &&
                                 chargeCheck &&
+                                (IsEnabled(CustomComboPreset.NIN_ST_AdvancedMode_Suiton_Uptime) && GetRemainingCharges(Ten) == 2 || (GetRemainingCharges(Ten) == 1 && GetCooldownChargeRemainingTime(Ten) < 3)) &&
```
Taurenkey, line 378: Remove this, you only need the code from L245-247 to enable the uptime feature. This also interferes with people that have the feature enabled to save 1 charge.

XIVSlothCombo/Combos/CustomComboPreset.cs
```diff
@@ -1863,12 +1863,16 @@ public enum CustomComboPreset
         NIN_HuraijinArmorCrush = 10064,
 
         [ParentCombo(CustomComboPreset.NIN_ST_AdvancedMode_RangedUptime)]
-        [CustomComboInfo("Raiton Uptime Option", "Adds Raiton as an uptime feature.", NIN.JobID)]
-        NIN_ST_AdvancedMode_Raiton_Uptime = 10065,
+        [CustomComboInfo("Phantom Kamaitachi Uptime Option", "Adds Phantom Kamaitachi as an uptime feature.", NIN.JobID)]
```
Taurenkey, line 1866: Please use my original numbering scheme.
Raiton Uptime -> 10065
Phantom Uptime -> 10066,
Suiton Uptime -> 10067

Taurenkey (changes requested):

XIVSlothCombo/Combos/PvE/NIN.cs
Daemeous: Without that check, the uptime feature won't function. The Suiton needs to be delayed rather than fired off immediately, as per the description to the new Suiton uptime. It doesn't interfere with ChargeHold, it actually uses the very same check as it.

XIVSlothCombo/Combos/CustomComboPreset.cs
Daemeous: I'm happy to do that, but is there a specific reason for this? Having it ordered in priority terms and then having the IDs updated to match made sense to me. Do you want me to restore the previous order too?

Taurenkey: We never change numbers once they're live, unless it's a full reset which this does not call for. These dictate people's settings, arbitrarily changing numbers isn't a fun time to explain to people why stuff is turned off and something else is turned on.

Daemeous: And these aren't live so now is the time to change them if we want to, no?

XIVSlothCombo/Combos/PvE/NIN.cs
Taurenkey: It will work once you add in `IsNotEnabled(CustomComboPreset.NIN_ST_AdvancedMode_Suiton_Uptime` to L191

Daemeous: Let me clarify: it does technically work but Suiton normally is fired off immediately once the conditions are met. The feature without this delay is pointless, the delay there is to hold it until we're about to cap. Though thinking about it I should also add one for if trick is about to come off cooldown, which also does ChargeCheck, so that Trick doesn't risk drifting without ChargeHold on...but that's already a monstrous if statement line...

Taurenkey:
Closing as #1091 now subercedes this.

PR #1042, opened 2022-10-09
[SGE] Add SGE_ST_Toxikon

jundong321:
* SGE_ST_Toxikon: group spells with no cast time in one button, so that you can spam it when you are moving.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -2367,8 +2367,8 @@ namespace XIVSlothCombo.Combos
             SGE_ST_Toxikon_Lucid = 141110,
 
             [ParentCombo(SGE_ST_Toxikon)]
-            [CustomComboInfo("Eukrasian Dosis Option", "Use Eukrasian Dosis if DoT is needed or nothing else can be used.", SGE.JobID, 1120, "", "")]
-            SGE_ST_Toxikon_EDosis = 141120,
+            [CustomComboInfo("Eukrasian Dosis Uptime Option", "Use Eukrasian Dosis for DoT uptime.", SGE.JobID, 1120, "", "")]
+            SGE_ST_Toxikon_EDosis_Uptime = 141120,
 
             [ParentCombo(SGE_ST_Toxikon)]
             [CustomComboInfo("Phlegma Option", "Use Phlegma if available and within range.", SGE.JobID, 1130, "", "")]
@@ -2377,6 +2377,10 @@ namespace XIVSlothCombo.Combos
             [ParentCombo(SGE_ST_Toxikon)]
             [CustomComboInfo("Dyskrasia Option", "Use Dyskrasia if Toxikon is not available and within range.", SGE.JobID, 1140, "", "")]
             SGE_ST_Toxikon_Dyskrasia  = 141140,
+
+            [ParentCombo(SGE_ST_Toxikon)]
+            [CustomComboInfo("Eukrasian Dosis Option", "Use Eukrasian Dosis if no other skills can be used.", SGE.JobID, 1150, "", "")]
+            SGE_ST_Toxikon_EDosis = 141150,
             
             #endregion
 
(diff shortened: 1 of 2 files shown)


Daemeous:
I believe the AoE DPS feature already achieves this functionaliuty.

Daemeous:
Dummy hitboxes can funky you might want to instead test on something your chocobo is tanking. I'll take a look at this tomorrow though. I've been toying with the idea of writing a more primitive version of this myself for a while now.

jundong321:
Yeah I've tested the distance thing with p7n boss, but feel free to test again.

Daemeous:
Perhaps you should consider clarifying the description so it indicates it defaults to spamming Eukrasian Dosis when it can't do anything else.

jundong321:
> Perhaps you should consider clarifying the description so it indicates it defaults to spamming Eukrasian Dosis when it can't do anything else.

Do you want it to be more descriptive than "Use Eukrasian Dosis if DoT is needed or nothing else can be used."? Open to the option.

Daemeous:
Ah, I misread. Honestly, if it were up to me, I'd split that out into two options. One for refreshing and one for if nothing else is available. That way it could double as an AoE button that spams Eukrasian Dosis if you step out of range. Good for expert roulettes for keeping Kardia ticking.

jundong321:
Sounds good, changes are made and tested on my machine

Daemeous:
> I maybe reading this wrong, but isn't this just the existing Phlegma Combo with DoT?

And Dyskrasia if you're in range for it, essentially it's a "trigger the Kardia whilst moving no matter what whilst doing the most dps possible" button

jundong321:
> I maybe reading this wrong, but isn't this just the existing Phlegma Combo with DoT?

Like Daemeous said, this is a feature for maintaining GCD uptime when you are moving. Phlegma Combo with DoT is for aoe IIRC.

Tartarga:
Took another look again, also very similar to single target DPS, and that you're mentioning just uptime damage.

Would this tiny revamp of single target dps work as good?

image

jundong321:
How do you use these skills without moving then?

> Add a bulleted list, <Ctrl+Shift+8>Add a numbered list, <Ctrl+Shift+7>Add a task list, <Ctrl+Shift+l>
> Directly mention a user or team
> Reference an issue, pull request, or discussion
> Add saved reply

Daemeous:
> Took another look again, also very similar to single target DPS, and that you're mentioning just uptime damage.
> 
> Would this tiny revamp of single target dps work as well?
> 
> image

The weakness there is you pick only one and either Toxicon runs out of charges and it gets stuck or if you run out of range for Dyskrasia you dyskrasia nothing and burn mana on 0 damage/healing

Tartarga:
> > Took another look again, also very similar to single target DPS, and that you're mentioning just uptime damage.
> > Would this tiny revamp of single target dps work as well?
> > image
> 
> The weakness there is you pick only one and either Toxicon runs out of charges and it gets stuck or if you run out of range for Dyskrasia you dyskrasia nothing and burn mana on 0 damage/healing

I already thought of that. It's not a binary choice that I rigged up
image

Daemeous:
Then it would have to also default to using Eukrasian Dosis once both of those fail to work and you're moving.

jundong321:
> > > Took another look again, also very similar to single target DPS, and that you're mentioning just uptime damage.
> > > Would this tiny revamp of single target dps work as well?
> > > image
> > 
> > 
> > The weakness there is you pick only one and either Toxicon runs out of charges and it gets stuck or if you run out of range for Dyskrasia you dyskrasia nothing and burn mana on 0 damage/healing
> 
> I already thought of that. It's not a binary choice that I rigged up image

I'm sure you can make the feature work exactly the same in there, just under movement condition.
But it's pretty common for people to intentionally use a non-casting skills to either prepare to move or double weave. Having to move in these cases won't work nicely.
If you want simplicity, I can remove all the suboptions, and leave only the top option that enables all suboptions there. And more description in the option to describe what it is.
So we will end up with a simple option that turns toxikon into a button where you can spam for movement or double weave. WDYT?

Daemeous:
I like the way it works now, we get that functionality of your last sentence just turning everything on. Being able to use this button as a "gotta double weave now" is also a great benefit, not just full mobility.

PR #1047, opened 2022-10-10
Dancer Devilment Priority

Xe-Xo:
Standard Step was being prioritized over Devilment in the edge case of Standard Step coming off cooldown while Technical Dancing. 

This is a huge DPS overall and a relatively simple priority fix.


k-kz:
Closing this - addressed in #1058 alongside other stuff

PR #1048, opened 2022-10-10
SCH ED Slider

Tartarga:
SCH Energy Drain slider moved under an advanced option similar to Bio as slider defaults don't exist yet.

Slider max now goes up to 60 seconds


Tartarga:
Holding while investigating Taurenkey's suggestion

k-kz (approved):
Changed descriptions as discussed, rest looks good to me 👍🏼

PR #1049, opened 2022-10-10
Advanced DoT menus on AST/SGE/WHM

Tartarga:
DoT menus on the other 3 healers
Region'd SGE's Config
Added Green + on those Dot Menus


Tartarga:
Before Alignment Edit
image

After Alignment Edit
image

PR #1051, opened 2022-10-12
UserConfig Separation

Tartarga:
UserConfig split into 2 files. New file in Combos (UserConfigItems) contains only the end user UI, where UserConfig retains the framework. Git magic done so UserConfigItems has git history attached

image
image


Tartarga:
If this needs to wait for any other projects that are in mid progress, I can recreate the split at any time and resubmit.

PR #1066, opened 2022-10-16
[MNK] internal cleanup + first try at Simple Single Target rotation

Kagekazu:


Kagekazu:
did some internal cleanup of the levelchecks and added the Simple Target (might need another pair of eyes to check if everything is working on Sloth Standard).

k-kz:
#1077

PR #1070, opened 2022-10-17
[MNK] added Simple AOE, ADV AOE and basic AOE

Kagekazu:
divided AOE like ST,

Reordered UI (prob will have to recheck all options u had active on the UI for MNK, since i'm UI scrub)


k-kz:
#1077

PR #1084, opened 2022-10-25
[DRG] Disembowel and chaos thrust changes

Kagekazu:
- Changed Disembowel and Chaos Thrust so its usable on more than 1 target.
- Changed start skill to 1st in chain - True Thrust - instead of 3rd in chain - Full Thrust - .

(works on lvl 90, but need some extra confirmation for below lvl 86 if possible)


Kagekazu:
Same as #1083. Will close this one

PR #1089, opened 2022-10-30
[SCH] Cleanup

Tartarga:
(holding till prereqs are done)
- Single Target DPS
  - Redesigned Alt mode to match Sage's Advanced Options.
    - Options redone similarly to Reaper's SoulSow (select actions to bind feature to)
    - Bio's action option no longer "idles" as Broil
    - Ruin II added as an option
- Removed Feature "Bio on Ruin II"
- Aetherflow Feature
  - UI Updates. Excogitation and Indominability now are user options, not preset options
- Deployment Tactics Feature: Code cleanup to use new GetHealTarget code with optional Mouse over checking


jundong321:
Is "Bio on Ruin II" removed or added to single target DPS mode?

Tartarga:
> Is "Bio on Ruin II" removed or added to single target DPS mode?

It's added to Single Target

PR #1102, opened 2022-11-06
Draft BLM: Opener and Rotation Update

thunderebolt:
- Adds a better opener using double transpose weave
- Adds a transpose line every minute swiftcast is up
- Fixes a minor bug with Blizzard3 at lower levels
- Allows spamming of umbral soul when there's no target
- Fixes Formatting in `CustomComboPreset.cs` File

### Only for Kage's reference as he adds this as an advanced BLM option!


k-kz:
Closing due to replacement via #1109

PR #1151, opened 2023-01-13
[RPR] Add TargetNeedsPositionals() logic

dlm21:
Add protection to Reaper from using True North when target doesn't need it


dlm21:
Was revealed that an ealrier PR already handles this almost exactly the same way, https://github.com/Nik-Potokar/XIVSlothCombo/pull/1057 , plus adds 2 other classes, so it's probably the better option

k-kz:
Closing as per previous comment - will implement via #1057

PR #1251, opened 2023-06-21
[Framework] Update `ShieldValue`

k-kz:
Apparently `Character.ShieldValue` became obsolete at some point


Tartarga:
Wished I knew more on stucts to give a yay/nay but it appears fine?

Issue #1257, opened 2023-07-04
Very bad RDM pvp bug!


MikeOrtizzz:
Little too specific but I have notice and tested it a little bit

Wen you play redmage in any kind of pvp area and you get silenced all pvp actions transform into the pve action "Riposte" making you unable to use the pvp action "Purify" to get rid of the silence effect

I don't use any pvp features the pluging has and barely use any pve features, I do use a lot of the gathering ones. As for RDM I use the melee combo pve feature.

This issue seems to stop when I disable the plugin.

Tartarga:
Fix merged and released

PR #1265, opened 2023-07-28
"Riposte Leak" Fix

Tartarga:
Removed Jolt fallback to Riposte. Sloth at level 1 is pointless


Taurenkey:
Fixes #1257

Tartarga:
Ran crystaline twice, no issue on multiple silence/stuns

PR #1275, opened 2023-08-28
VS Warnings Cleanup

Tartarga:
Disabled unused var in BLU
Applied Obsolete recommendations to SCH/SMN Pets and Chocobo detection


Tartarga:
Covered in Taurenkeys API9 commit

PR #1287, opened 2023-10-04
Update XIVSlothCombo.csproj

ChrisMellor:
DalamudPackager needed updating, possibly the reason for the failed build???


Taurenkey:
Nah, reason was just it wasn't building against the staging version, but I'll merge this all the same because I'm too lazy to go into the Nuget manager to update

PR #1291, opened 2023-10-13
RPR and DRG changes

levinkin:
- adds option to separate harvest moon in "Ranged Filler Option" for RPR 
- fixes "Dives Options" for DRG
  - prevent Stardiver from clipping by changing the order of things
  - delays Stardiver if high jump is just about to go off cd to prevent drifting of high jump
- optimizes "Wyrmwind Thrust Option" for DRG 
  - prioritizes buffs over wyrmwind so that wyrmwind will be used with buffs when used with "Buffs on Main Combo" options
- optimizes "Life Surge Option" for DRG
  - prioritizes use of life surge over other ogcds so that it actually gets used when proper gcd is about to come up while making sure geirshkogul and jump doesnt drift
  - fixed life surge not being used on fang and claw
  - makes it so that life surge is being used on optimal times based on balance discord info (https://discord.com/channels/277897135515762698/582430670006124564/918555147347587082) 
- made it so that blood bath can be used as an opener indicator instead of true north since left eye does the same thing anyway
- optimizes jumps so that it doesnt clip anymore

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
       uses: microsoft/setup-msbuild@v1.0.2
     - name: Download Dalamud
       run: |
-        Invoke-WebRequest -Uri https://goatcorp.github.io/dalamud-distrib/stg/latest.zip -OutFile latest.zip
+        Invoke-WebRequest -Uri https://goatcorp.github.io/dalamud-distrib/latest.zip -OutFile latest.zip
         Expand-Archive -Force latest.zip "$env:AppData\XIVLauncher\addon\Hooks\dev\"   
     - name: Build
       run: |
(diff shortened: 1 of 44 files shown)


Taurenkey (changes requested):
Please undo your changes to the build.yml file and the README.md file.

Taurenkey:
You've made too many changes that no longer fit your PR. I won't be merging in this state. Please break up everything that wasn't part of the original PR intended changes either as another PR or don't submit them.

levinkin:
> You've made too many changes that no longer fit your PR. I won't be merging in this state. Please break up everything that wasn't part of the original PR intended changes either as another PR or don't submit them.

fixed and placing samurai optimizations in a separate pr

PR #1356, opened 2024-02-17
Force LifeSurge to only be used on 5th Combo GCD

Trutix-dev:
The 5th combo GCD in a Dragoon's rotation has 100 more Potency due to the Trait "Lance Mastery I".

While using Sloth I've noticed that outside of the Opener, it will use Life Surge on the 4th GCD instead of the 5th. Resulting in a huge Potency Loss overtime.

I added an extra check to DRG.cs that only uses LifeSurge on a positional if a positional has been done beforehand. 
  
Please review, I'm tired and I'm not sure if I placed all the ( ) correctly. :P


Kagekazu:
this became obselete with the DT changes of DRG

Issue #1385, opened 2024-06-03
[SUM] Egi summon are always blocked by Demi.


tex111:
Summoner (Advanced Summoner Feature)

Egi summons are blocked by a Demi being off cooldown even if the option that adds Demi summons to the combo is disabled.

Kagekazu:
thanks for letting us know! ill see to fix it in next release after Dawntrail!

Genesis-Nova:
Fixed with #1404.

PR #1386, opened 2024-06-03
Fix Egi summons always being blocked by Demi.

tex111:
Since the Egi summon is so low in the tree, the check for Demi summons being off cooldown isn't necessary and prevents players who choose not to add demi summons to the combo from summoning an egi when a demi is off cooldown.

This fixes https://github.com/Nik-Potokar/XIVSlothCombo/issues/1385


Genesis-Nova:
This can be closed, I rolled this fix in with my DT smn fixes.
Thank you for your contribution @tex111.

PR #1400, opened 2024-06-30
Update Ability and Status IDs

Dubsys:
New actionids and buffids

90 combo should work, higher level abilities not implimented but the ids are there.


SyniRon:
WHM is missing Sacred Sight, the new buff from POM

Dubsys:
> WHM is missing Sacred Sight, the new buff from POM
> 
> Should be 3879

Added, thanks.

itzeddie:
This doesn't actually build successfully for me. Looks like an error on line 36 (missing ,). And the buffs under PCT throw an error in the build too: "The name 'Starstruck' does not exist in the current context."

PR #1433, opened 2024-07-06
[PLD/GNB] Goring Blade all levels fix, Various GNB 7.0 fixes/adjustments

### LINKED ISSUE #1424: XIVSlothCombo 3.2.0.0 - [VPR]+[PLD] Issues
VPR: Reawaken is in "Advanced Mode - AoE" Rotation even if its disabled. 
PLD: "Goring Blade" in "Simple Mode - Single Target" and in "Advanced Mode - Single Target" is not spend if its activated.

Sorry for my bad English. :-)

### LINKED ISSUE #1451: [GNB] Advance AoE Sonic Break not working as purpose?
As one button Aoe Feature hava Sonic Break on AoE Feature it does use it. But after using it once it is still want to use it again even it is imposiable as it can only use once during No Mercy

### LINKED ISSUE #1453: [GNB Request] Fated Circle into Continuation (Fated Brand)
Currently, Continuation will appear on the Demon Slice AoE combo following Fated Brand.
Would it be possible for there to be an option to have the Fated Circle button be followed by Continuation (Fated Brand), like how Burst Strike has the same option?

### LINKED ISSUE #1454: [PLD] using intervene cancels out the new goring blade
as described, done the new extreme and noticed that a couple of times my goring blade did not go off during a burst window, I went to a dummy and found out that using intervention and possibly something else  cancels out the cast shown on the button that  replaces fast blade and moves to a next  step

### LINKED ISSUE #1461: [GNB PVP] Double charge
GNB PvP combo has you double charge on cooldown and at the start of combo, wastes charge.

### LINKED ISSUE #1470: [GNB] Double down Single Target
In advanced mode Single Target Double Down appears in the single target rotation but is not used.

### LINKED ISSUE #1471: [GNB] extremely small window to use No mercy
Using the [Gnashing Fang Continuation Combo Feature] with [No Mercy on Gnashing Fang Option] enabled you have less than a second to press No Mercy before it just changed back to the Gnashing Fang skill and won't show No Mercy again until you've competed another round of your 1 2 3 combo, then once again only changes to No Mercy for a split second.

Akechi-kun:
we got a live one
fixes #1424 
fixes #1451 
fixes #1453
fixes #1454 
fixes #1461
fixes #1470
fixes #1471

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1513,11 +1513,11 @@ namespace XIVSlothCombo.Combos
 
         #region Fated Circle
         [ReplaceSkill(GNB.FatedCircle)]
-        [CustomComboInfo("Fated Circle Features", "Collection of Fated Brand related features.", GNB.JobID)]
+        [CustomComboInfo("Fated Circle Features", "Collection of Fated Circle related features.", GNB.JobID)]
         GNB_FC = 7600,
 
         [ParentCombo(GNB_FC)]
-        [CustomComboInfo("Fated Circle Continuation Feature", "Adds Hypervelocity on Fated Circle.", GNB.JobID)]
+        [CustomComboInfo("Fated Circle Continuation Feature", "Adds Continuation on Fated Circle.", GNB.JobID)]
         GNB_FC_Continuation = 7601,
 
         [ParentCombo(GNB_FC)]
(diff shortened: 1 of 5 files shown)


Lusaca87:
One Issue: in advaned mode: if you have Confiteor and Blade Options Both disabled goring blade will not be used

Ceveth:
Some <90 logic for dumping charges during NM to keep Blood Fest aligned at 76+

PR #1437, opened 2024-07-06
Preliminary BRD changes

NebulousByte:
Preliminary adjustments. Might need some tweaks, but otherwise seems fine.
- Swapped Straight Shot Ready for Hawk's Eye and added additional Barrage proc
- Changed return Bloodletter to return OriginalHook(Bloodletter) for HeartbreakShot upgrade
- Added Resonant Arrow option
- Added Radiant Encore option

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -684,10 +684,6 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Simple Raging Jaws Option", "Enable the snapshotting of DoTs, within the remaining time of Raging Strikes below:", BRD.JobID)]
         BRD_Simple_RagingJaws = 3025,
 
-        [ParentCombo(BRD_Simple_DoT)]
-        [CustomComboInfo("Opener Only Option", "Until the first auto-refresh you can DoT new targets automatically.", BRD.JobID)]
-        BRD_Simple_DoTOpener = 3026,
-
         [ParentCombo(BRD_AoE_Simple_Songs)]
         [CustomComboInfo("Exclude Wanderer's Minuet Option", "Dont use Wanderer's Minuet.", BRD.JobID)]
         BRD_AoE_Simple_SongsExcludeWM = 3027,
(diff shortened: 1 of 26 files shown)


Connorbeaz:
Actions that are effected by Hawk's Eye (Straight Shot/Refulgent Arrow and Wide Volley/Shadowbite) arent being added into the Simple Bard Feature or Simple AOE Bard Feature

NebulousByte:
I've cleaned it up a bit and tested it. This is adding Straight Shot/Refulgent Arrow and Wide Volley/Shadowbite for me. I've also added Barrage to Simple AoE for the 270 Potency Shadowbite.

zbee:
DoTs in the single target rotation are not being maintained, and are only applied individually once.

NebulousByte:
It works fine for me. Can you give more details on the use case for this? e.g. Level/sync,, options, etc.?

zbee:
I was synced down to 80 in Puppets Bunker - It applied Caustic+Storm manually once, then it wouldn't use Iron Jaws so I had to apply both manually, then it would use Iron Jaws as expected after that.
It works in overworld unsynced stuff though.

NebulousByte:
I've tested in Paglth'an with Duty Support twice now and I can't replicate the issue. What options are you using? Are you using the Opener Only option?

zbee:
Ah yes, I am using the Opener Only option. Rereading it now ... is it just working as intended?
image

NebulousByte:
I'm unsure if it is working as intended. I haven't touched the original logic behind this, but it seems fine in every test I've done, both with Opener Only and without. The opener only option should only affect whether the DoTs are applied to multiple targets outside of the opener, not whether Iron Jaws is used? By my understanding, Iron Jaws should be used under two circumstances:

1. If you're still within 3 seconds of Raging Strikes
2. If the DoTs you've applied have less than 4 seconds duration

But from what you've said is that the plugin applies both DoTs and then does not use Iron Jaws until you use it at least once manually, thus causing the DoTs to run out and having to re-apply them manually.

zbee:
With Opener Only Option disabled the class appears to work perfectly 👍🏻

Scjaro:
I tested the Bard, it doesn't use Pitch Perfect in the rotation, but if you activate the "Action Change Settings" for "The Wanderer's Minuet" it works - is that intended? :D (Lv. 92)

NebulousByte:
Considering the code explicitly looks for songWanderer, I'd say it's working as intended. This wasn't something I changed, mind you. So in terms of intent, you'd need to ask the person who originally wrote the BRD code. I _can_ change it if deemed appropriate to do so.

PR #1483, opened 2024-07-09
GUI Scaling Bug in PvE Features Fix

DallagenXYZ:
image

Solves the issue pictured above, where raising the font scaling in Dalamud causes the margins in between the tab list and the scrollable region below to change and start clipping 

image

Also fixes the Job Icon's scaling and keeps it proportional to the text in the tab list title


DallagenXYZ:
same issue exists in pvp features for the same reason as well

DallagenXYZ:
i messed up my branch management will reopen with pvp features fixed

PR #1491, opened 2024-07-11
Ast

Taurenkey:


DallagenXYZ:
This resolves the icon loading but does not resolve the icons being improperly scaled with the dalamud font size which causes gui clipping

Taurenkey:
> This resolves the icon loading but does not resolve the icons being improperly scaled with the dalamud font size which causes gui clipping

It does. The icons X & Y are scaled according to Dalamud settings before any adjustments.

DallagenXYZ:
Apologies, local build issue, I cloned the branch but it didn't actually pull half the updated files the first time

PR #1497, opened 2024-07-11
WHM Support disabling of Action Change

zbee:
- [X] Make Divine Caress compatible with disabled Action Change on Temperance
- [X] Make Glare IV compatible with disabled Action Change on Presence of Mind


Taurenkey:
> * Make Divine Caress compatible with disabled Action Change on Temperance

Seems missing?

zbee:
Fixed now in 8c47ad9.

PR #1530, opened 2024-07-13
[GNB] Full ST Rotation Rework

### LINKED ISSUE #1313: [SUGGESTION] [GNB] Adding a 1s CD check to Gnashing Fang
There are certain instances where Gnashing Fang will come up right as my GCD & will cause GCD to be used instead of Gnashing Fang on a 30s window, causing No Mercy & Gnashing Fang to misalign. iirc there is a certain check on MCH where if you use Drill (i.e.) it comes up about 1s before it is used, pretty much guaranteeing that it is used instead of a 1-2-3. Is it possible to add this same check from MCH to GNB for Gnashing Fang? I feel this would benefit it entirely.

### LINKED ISSUE #1326: [BUG][GNB] Bug One Button Feature during the 'No Mercy' activation period (Sks 2.5s)
The moment you press 'No Mercy' and use 'Burst Strike' during 'Hypervelocity', it interferes when using 'Bloodfest' instead of pressing 'Double Down' when it's ready, but it turns out to be using Burst Strike x3 Times instead

### LINKED ISSUE #1564: [GNB] Lionheart Combo not working after 1st usage (opener)
On advanced mode, Lionheart combo will not be used after the 1st usage during the opener.

### LINKED ISSUE #1570: GNB: Reign combo on Main Combo Option / Burst Strike to Bloodfest Feature Problems
Unfortunately, both functions have a major problem: In the No Mercy Phase, the 3-piece combo is not displayed when activating Bloodfest. Neither with the Main Combo nor with the Burst Strike. 

Scenario 1: I connect Reign to the main combo and when activating Bloodfest it is not displayed in the main combo.

Scenario 2: Combining Burst Strike with Bloodfest is a good idea, but unfortunately poorly implemented, as you are not transferred to the Reign combo after spending the cartridges if you have no more cartridges for Burst Strike.

Akechi-kun:
fixes #1313
fixes #1326 
fixes #1564 
fixes #1570 
Changelog
- reworked entire rotation logic
- cleaned up subLv100 rotations
- fixed Lv90 rotation from 2cart 2min bursting (thank you Kage)
- adjusted actions to better fit with rotation & buff windows

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1422,7 +1422,7 @@ namespace XIVSlothCombo.Combos
 
         #region ST
         [ReplaceSkill(GNB.KeenEdge)]
-        [CustomComboInfo("Advanced Gunbreaker Feature", "Replace Keen Edge with its combo chain and uses Burst Strike to prevent ammo overcap. ***DOES NOT WORK WELL WITH 2.46-2.49***", GNB.JobID)]
+        [CustomComboInfo("Advanced Gunbreaker Feature", "Replace Keen Edge with its combo chain and uses Burst Strike to prevent ammo overcap. **2.5sks ONLY**", GNB.JobID)]
         GNB_ST_MainCombo = 7001,
 
         #region Reign Combo
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1548,6 +1548,10 @@ namespace XIVSlothCombo.Combos
         GNB_BS_Bloodfest = 7402,
 
         [ParentCombo(GNB_BS)]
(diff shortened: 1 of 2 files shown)


XIVSlothCombo/Combos/CustomComboPreset.cs
```diff
@@ -1547,6 +1547,10 @@ public enum CustomComboPreset
         [CustomComboInfo("Burst Strike to Bloodfest Feature", "Replace Burst Strike with Bloodfest if you have no powder gauge.", GNB.JobID)]
         GNB_BS_Bloodfest = 7402,
 
+        [ParentCombo(GNB_BS)]
+        [CustomComboInfo("Reign combo on Burst Strike Feature", "Adds Reign combo to Burst Strike when under No Mercy and when Double Down & Gnashing Fang are on CD.", GNB.JobID)]
+        GNB_BS_Reign = 7403,
```
Taurenkey, line 1552: Same ID as line 1556.

Akechi-kun: Mistake on my end. Sorry about that. Was a last minute add in, but no excuses.

PR #1533, opened 2024-07-13
MCH Bugfixes

### LINKED ISSUE #1092: [MCH] `Wildfire`/`Hypercharge` at HP% option
Hello.

im unsure if i do it in the right order, and hope you can understand me.
With my handicap someone shows me this plugin and i try it yester day.
it works simple but i realized in the rotation a little problem

after an internal try with ucob the fixed rotation used wildfire everytime ab 7% HP, and with the intermission to the next phase you will feed it with 2-4 weaponskills and its on CD. for the next boss where you want it ready. it would be nice if there is a way like for DNC to prevent using this skills at % hp bar.

i disabled hypercharge and wildfire atm, but i want to let you know.

best regards

### LINKED ISSUE #1381: Crash with sloth config open. Also FPS weirdness.
I've noticed this in the past as well where the only time Sloth will crash my game is if the configuration window is open.

I wasn't even in combat and was just running around clicking config settings (Machinist ones if that matters), and the game crashed. This has happened in the past, too.

I haven't had any other crashes in this game for a very long time now so I doubt it's something else.

Another issue is that just having the /scombo window open will drop my FPS down by 10 frames, which is odd.

### LINKED ISSUE #1504: [MCH} Wildfire and Heat Guage
Wildfire issue still persists, doesnt use it after 2 min window. Also refuses to use heat guage past 2:30. Overcaps Heat Guage by A LOT (I assume its cause of the WF issue). Like 2:30 - 4:00ish doesnt use any heat guage.

This was tested in both Everkeep(EX) and Lv 100 Dummy (Stone, Sky, Sea Dummy in solution 9)
Tried both Advanced and Simple.

Thanks as always for the hard work <3

### LINKED ISSUE #1519: [MCH] Issue
image
image

### LINKED ISSUE #1534: [MCH] Single Button Heat Blast Feature only works with Heat Blast and not Blazing Shot
As the title suggests, when enabling the Single Button Heat Blast Feature on MCH, Blazing Shot (the upgrade of Heat Blast) is not replaced with Hypercharge (when above 50 Heat Gauge ofc). Blazing Shot is a potency upgrade over Heat Blast and it isnt even possible to assign to a hotbar after getting Blazing Shot (except for macros I guess...).

mch

Kagekazu:
- fixes WF issue on lvl 100
- fixes 2 min burst
- fixes overcap
- ! changed reassemble order - ppl will have to uncheck and recheck the options they want to have it on !
- various little things

fixes #1381 
fixes #1519 
fixes #1504 
fixes #1534 
fixes #1092

### DIFF
--- a/XIVSlothCombo/Combos/PvE/MCH.cs
+++ b/XIVSlothCombo/Combos/PvE/MCH.cs
@@ -328,14 +328,10 @@ namespace XIVSlothCombo.Combos.PvE
                     if (IsEnabled(CustomComboPreset.MCH_ST_Adv_Interrupt) && interruptReady)
                         return All.HeadGraze;
 
-                    if (IsEnabled(CustomComboPreset.MCH_ST_Adv_QueenOverdrive) && gauge.IsRobotActive && GetTargetHPPercent() <= Config.MCH_ST_QueenOverDrive &&
+                    if (IsEnabled(CustomComboPreset.MCH_ST_Adv_QueenOverdrive) &&
+                        gauge.IsRobotActive && GetTargetHPPercent() <= Config.MCH_ST_QueenOverDrive &&
                         CanWeave(actionID) && ActionReady(OriginalHook(RookOverdrive)))
                         return OriginalHook(RookOverdrive);
-
-                    // BarrelStabilizer
-                    if (IsEnabled(CustomComboPreset.MCH_ST_Adv_Stabilizer) &&
-                        !gauge.IsOverheated && CanWeave(actionID) && ActionReady(BarrelStabilizer))
-                        return BarrelStabilizer;
 
                     if (IsEnabled(CustomComboPreset.MCH_ST_Adv_Hypercharge) &&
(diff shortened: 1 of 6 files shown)


XIVSlothCombo/Combos/PvE/MCH.cs
```diff
@@ -697,7 +733,7 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                     if (IsEnabled(CustomComboPreset.MCH_Heatblast_Wildfire) &&
                         ActionReady(Hypercharge) &&
```
Evil-Crab, line 734: Is this check necessary? As far as I can tell checking for both `ActionReady(Hypercharge) && WasLastAbility(Hypercharge)` will always return false. I had to remove `ActionReady(Hypercharge)` in my build to make this part work,

Kagekazu: should be levelcheck instead of actionready.

PR #1543, opened 2024-07-14
VPR Bugfixes

### LINKED ISSUE #1490: Huge fps drop on SlothCombo usage
I decided try VPR PVE features and i opened nearly all options but i noticed the moment i get in combat and spam button my fps drops from 100+ to 25 fps. I assume it is happening because i open too much option but it is still weird cause i have already pretty strong pc. i am adding debug file also. I tried to close all options still fps dropping to 25 the moment i close plugin everything turns back to normal
[SlothDebug.txt](https://github.com/user-attachments/files/16167923/SlothDebug.txt)

### LINKED ISSUE #1506: [PLD] + [VPR] + [all melee?]
PLD: have a checkbox to prevent Intervene (the gap closer) being used when character is moving.

encountered couple times where right at Confiteor range combo, I tried to move out of Boss' aoe but due Intervene beeing pooled for burst, it kept dashing right back to the boss.

i dont mean the "only use when in melee" this one works fine, its more like greeding 1 last GCD before moving out and its still considered melee range, tho running out. then Intervene will launch us back into the hit box and point blank aoe.
____________________
VPR: have the odd minute reawaken be placed at the 1 minute mark, rather than right as it comes up. since this one is flexible, this could be held for potential dps gain under AST cards or other 1 minute buffs. or maybe even have a specifically checked if AST card buff is present -> go into reawaken (with checking to not drop personal buffs ofc)

VPR: Noxious Gnash debuff is falling off during double reawaken, even with Gnash timer set to maximum of 20sec.
____________________
All Melee?: additional checkbox for the opener only: to use True North before positional attacks. i notice often on VPR when im about to hit positionals, tanks would still be positioning the boss which makes positionals "sometimes" goes wild. i tend to use True North myself mostly for convinience, but could be a feature within Sloth maybe?

### LINKED ISSUE #1510: [Suggestion] [VPR] True North after Dread Winder/Pit of Dread
Whenever Dread Winder/Pit of Dread is used, True North is immediately used in preparation for the Twinblade combos.

OR alternatively

True North is automatically used before Dread Winder/Pit of Dread as an option

### LINKED ISSUE #1514: [VPR] 2 minutes
I noticed when testing the changes that the double reawaken windows at 2/4/6 minute windows  noxious gnash will be falling off as the 2nd reawaken starts. other than that it works beautifully!

### LINKED ISSUE #1516: [VPR] small error during opener
i9ap3ED

in Sloth opener the actions 12 and 14 are swaped compared to the balance opener.
After reawaken Sloth is doing 2x uncoild fury, rather than the optimal hindsting strike and THEN uncoild fury.

if i press Hindsting Strike manually after reawaken, Sloth is locking himself from continueing and next skill is greyed out

### LINKED ISSUE #1527: [VPR] Reawaken issues, combo finishers issues
1. The debuff sometimes runs out before double reawaken,
2. Reawaken isn't used on even/odd minutes after 2 min+, it's usually used when the cooldown on Serpent's Ire is at 70+ seconds, which makes everything not align for later 2 minute bursts,
3. Idk if google docs links are allowed here, but ill try to post one here. After making a copy, the doc shows a 10 minute rotation that could be implemented, at 2.50s GCD, during full uptime. It loops perfectly up until then. I don't know if the devs can be bothered with changing it to this, but it would be nice.
4. There is a specific moment in rotation where it skips a Death Rattle and goes straight into Uncoiled Fury,
5. There are issues with executing some combo finishers where there is just not enough time to execute it and the time on the extra potency finisher runs out.

https://docs.google.com/spreadsheets/d/18haZwiTdqaNaKK3P6_n_i7uFp10c-cDT696owUaETPc/edit?gid=1149453378#gid=1149453378

### LINKED ISSUE #1540: [VPR] Attack actions disabled when Noxious Gnash abilities are not learned yet
This issue affects at least Simple mode Single Target, Advanced Mode Single Target, and Advanced Mode AoE, but based on what I've seen, i'm pretty sure Simple Mode -AoE will be impacted as well.

This issue is when Dread Fangs is not known yet (character level < 10) or Dread Maw is not learned (character level < 35), a disabled Dread Fangs/Dread Maw is offered instead of Steel Fangs/Steel Maw.  This makes these features completely non-functional < lvl 10 for single target and < lvl 35 for AoE.  In the advanced modes, you can work around the issue by removing the Noxious Gnash check in both advanced modes, but it is always present in simple mode.

This is problematic in PotD solo runs while in the first 2 PotD floor sets.

### LINKED ISSUE #1541: [VPR]  Swiftskin's Sting not correctly offered below level 30
Before the major combo finishers are learned at lvl 30, Advanced Mode-Single Target does not seem to offer Swiftskin's sting reliably and the automation gets stuck always offering Hunter's string instead at that phase of the combo.  The expectation is that Hunter's Sting/Swiftskin's sting is prioritized based on which associated buff has the least amount of time remaining on it. 

The logic does not appear to be working correctly when the Fang/Strike abilities are not learned yet.  This is easiest to replicate on PotD floor sets 11-20.

Kagekazu:
- fix 2 min not always under noxious gnash
- add TN to dreadwinder
- cleanup

fixes #1510 
fixes #1514 
fixes #1516 
fixes #1540 
fixes #1527 
fixes #1490 
fixes #1541 
fixes #1506

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3795,7 +3795,7 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Uncoiled - Twins", "Replaces Uncoiled Fury with Uncoiled Twinfang and Uncoiled Twinblood.", VPR.JobID)]
         VPR_UncoiledTwins = 30202,
 
-        [ReplaceSkill(VPR.Reawaken)]
+        [ReplaceSkill(VPR.Reawaken, VPR.DreadFangs)]
         [CustomComboInfo("Reawaken - Generation", "Replaces Option with the Generations.", VPR.JobID)]
         VPR_ReawakenLegacy = 30203,
 
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3638,6 +3638,7 @@ namespace XIVSlothCombo.Combos
         #endregion
 
         #region VIPER
+
         [ReplaceSkill(VPR.SteelFangs)]
         [ConflictingCombos(VPR_ST_AdvancedMode)]
         [CustomComboInfo("Simple Mode - Single Target", "Replaces Steel Fangs with a full one-button single target rotation.\nThis is the ideal option for newcomers to the job.", VPR.JobID)]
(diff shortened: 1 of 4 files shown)


Scjaro:
image

Opener is fine.

PR #1412, opened 2024-07-03
DRK level 100

zbee:
With 7.0 DRK is a bit different overall, mostly due to a shift in how Delirium+Blood weapon stacks an timer are used.

- [X] Enable DRK
- [X] Update Action IDs
- [X] Add New Action IDs
- [X] Rework Combo options to make them more consistent with each other
- [X] Add option to ST>Living Dead to prevent it based on target %remainingHP (similar WHM option [here](https://github.com/Nik-Potokar/XIVSlothCombo/blob/06abee27d81e482c90b2c9c736e8e5db07ff1f94/XIVSlothCombo/Combos/PvE/WHM.cs#L281))
- [X] Add _separate_ option to AoE>Living Dead to prevent it based on target %remainingHP
- [X] Un-break Combo code
- [X] Update Combo code to 7.0 changes (no Blood Weapon, etc)
- [X] Add options for Disesteem use
- [X] Add level 100 actions into Combo code
- [X] Add new options into Combo code
- [X] Reorganize oGCDs in the Combo code to try to mitigate false positive Dark Arts reports preventing all other oGCDs from working
- [X] Organize Combo code to hopefully put out The Balance's [Standard Opener](https://discord.com/channels/277897135515762698/277962421862268930/1257508528357376060)
- [X] ~~Rework Combo Code to make it more consistent between combos~~
- [X] ~~Rework Delirium+Mana usage to build Mana on Odd minute windows~~
- [X] ~~Rework Combo options+code to reflect the new 'Simple' setup other classes are getting~~
- [X] ~~Update PVP for DRK~~
- [X] ~~Wipe existing configs (DNC did this [here](https://github.com/Nik-Potokar/XIVSlothCombo/pull/1266/commits/342996f3b9c53038938dafeee5e9a9805413baeb))~~

> [!caution]
> #### Inconsistent without ClientStructs+Dalamud updating DRK job gauge.
> Does not respect the user's chosen MP limit under Option#12 in some situations.
>
> <sub>No update should be needed here once those are updated and version bumped.</sub>

PR #1557, opened 2024-07-18
[Healers & PCT] Bug fixes and new utility features

StackBLU:
[WHM]
- Removed Cure 3 MP slider
- Fixed Cure 3 never being used for Simple AoE Heals (would always use Medica 1 instead)
- Fixed swift/thin air/raise feature (would get stuck on swiftcast before)
- Raised priority of Afflatus Solace above Regen (unlike aspected benefic, there is no immediate heal from regen - this logic will likely need to be adapted later on, as now regen is only used if you have no lilies)
- Added Misery check to Simple STHeals (allowed overcapping before)

[AST]
- Added lightspeed protection option
- Re-ordered IDs and UI ordering, as it didn't make much sense before

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3356,10 +3356,14 @@ namespace XIVSlothCombo.Combos
             SCH_DPS_EnergyDrain = 16005,
 
                 [ParentCombo(SCH_DPS_EnergyDrain)]
-                [CustomComboInfo("Energy Drain Burst", "Holds Energy Drain when Chain Stratagem is ready or has less than 10 seconds cooldown remaining", SCH.JobID)]
+                [CustomComboInfo("Energy Drain Burst", "Holds Energy Drain when Chain Stratagem is ready or has less than 10 seconds cooldown remaining", SCH.JobID, 1)]
                 SCH_DPS_EnergyDrain_BurstSaver = 16006,
 
-            [ParentCombo(SCH_DPS)]
+                [ParentCombo(SCH_DPS_EnergyDrain)]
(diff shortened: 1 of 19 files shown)


StackBLU (commented):
Sorry for the mess of commits - still learning how this works. I think I have everything that I actually wanted to upload good to go now though.

zbee:
You should reset AST options or not rearrange all of its options' IDs, it misinterprets all of my existing settings.

DNC previously did a reset of options [here](https://github.com/Nik-Potokar/XIVSlothCombo/pull/1266/commits/342996f3b9c53038938dafeee5e9a9805413baeb), which I was going to go off of, but Tartarga mentioned on #1412 that all configs were wiped for 7.0; they may have more information on how to do this correctly.

Taurenkey:
Rejecting this PR.

Starting with PCT, you've removed options altogether but added Lucid (which I have yet to find a use for with the job since it has no rez utility meaning MP is less of a factor for the job). You removed the combined aetherhues feature which is a no go since not everyone uses the full one button rotation but like having reduced button bloat.

Next, by reordering a lot of things you've made this review much harder to determine what's actually new and what's been removed, so rather than risk more options being removed because of hard to read diff checks, I'm rejecting it on this basis too.

Lastly, resets don't just get thrown out willy-nilly, there needs to be a significant reason for it. There was nothing here that supports the need for a reset.

StackBLU:
Combined aetherhues was not removed, it was split into advanced mode ST and
AoE like all the other classes. Same function, two checkboxes now instead
of one (which allows for more customizability in the future when adv mode
gets fleshed out).

As for the reorganizing of everything in healers, this was something I felt
was very much needed. There are inconsistencies between all of them. This
PR would have aligned all of them to follow the same pattern, which would
make developing easier.

If you want all the changes without the cleanup, I can do another one.

PR #1569, opened 2024-07-20
WHM Divine benison overcap protection

edewen:
Added DB overcap protection at 2 charges to the single target and holy spam combos. 
ST will weave easily. 
Aoe will weave when moving or typically when rapture overcap happens because holy spam itself doesn't give a weave window. 

Will only work for 88 and above as it is designed to retain one charge at all times to not get too automatey.


Taurenkey:
Doesn't really suit for the DPS options, since it doesn't do damage or aid in doing damage. Happy to accept an overcap protection for just the single target healing feature.

edewen:
Does this mean paladin will be losing its option of the exact same thing?

PR #1590, opened 2024-07-23
[PLD] Post Rework MP Reserve Feature

AustinSumigray:
Allows for only spending MP on skills when above a set number. The intended use case is reserving a bit of MP in case of emergency Clemency heals. The triggering pain point is a scenario where you have spent all your MP, and wish to be able to use Clemency as needed, but the combo wishes to spend it on Holy Spirit stacks or similar. You then can't use the combo itself to regain MP. This feature would skip casts of MP-using skills until you had built back up to the reserve value.

### DIFF
--- a/XIVSlothCombo/Attributes/CustomComboInfoAttribute.cs
+++ b/XIVSlothCombo/Attributes/CustomComboInfoAttribute.cs
@@ -51,7 +51,7 @@ namespace XIVSlothCombo.Attributes
 
         public uint ClassJobCategory => JobIDToClassJobCategory(JobID);
 
-        private int JobIDToRole(byte jobID)
+        private static int JobIDToRole(byte jobID)
         {
             if (Svc.Data.GetExcelSheet<ClassJob>().HasRow(jobID))
                 return Svc.Data.GetExcelSheet<ClassJob>().GetRow(jobID).Role;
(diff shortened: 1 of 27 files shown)


Akechi-kun:
This is very unlikely to be merged. This project doesn't support helping with mitigations or situational issues, it's made to execute a DPS rotation with either 1 button or reduce button bloat to any standards. This PR is just opening the door to start adding in mitigations, and then what's next that follows if this merges is everyone asking for the other Tanks' mitigations to be added, which isn't what this project is aimed for in the first place.

Taurenkey (changes requested):
This needs a bit of a tweak. Create seperate options for the ST & AoE rather than put them as just one unrelated option.

AustinSumigray:
> This needs a bit of a tweak. Create seperate options for the ST & AoE rather than put them as just one unrelated option.

Sure, there you go.

AustinSumigray:
> This is very unlikely to be merged. This project doesn't support helping with mitigations or situational issues, it's made to execute a DPS rotation with either 1 button or reduce button bloat to any standards. This PR is just opening the door to start adding in mitigations, and then what's next that follows if this merges is everyone asking for the other Tanks' mitigations to be added, which isn't what this project is aimed for in the first place.

The option doesn't use your mitigation for you, it just makes it so you can set a limit to how much of the resource you use for it you use on other tasks. A similar option is the Sheltron Option, that lets you set an amount of Oath gauge you want to have before you spend it on Sheltron.

Taurenkey (approved):

PR #1594, opened 2024-07-23
[DRG] fixes/adjustments

### LINKED ISSUE #1599: DRG uses twice heaven thrust then uses chaotic spring
DRG uses twice heaven thrust then uses chaotic spring

### LINKED ISSUE #1603: [DRG] Sometimes chaotic spring is triggered twice in a row.
thinking that I'm looking at the cooldown of LanceCharge, but there are times when I use chaotic spring twice in a row during a combo.

Akechi-kun:
fixes #1599 
fixes #1603 

### Changelog:
- Fixed PowerSurge/ChaoticSpring alignment issue; mistake on my end, left that line in there doing some testing

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRG.cs
+++ b/XIVSlothCombo/Combos/PvE/DRG.cs
@@ -206,10 +206,7 @@ namespace XIVSlothCombo.Combos.PvE
                             return (LevelChecked(OriginalHook(Disembowel)) &&
                                 (ChaosDoTDebuff is null ||
                                 ChaosDoTDebuff.RemainingTime < GCD * 5 ||
-                                GetBuffRemainingTime(Buffs.PowerSurge) < GCD * 7 ||
-                                GetCooldownRemainingTime(LanceCharge) < GCD * 4)) ||
-                                (!LevelChecked(ChaosThrust) &&
-                                GetBuffRemainingTime(Buffs.PowerSurge) < GCD * 4)
+                                GetBuffRemainingTime(Buffs.PowerSurge) < GCD * 7))
                                 ? OriginalHook(Disembowel)
                                 : OriginalHook(VorpalThrust);
                         }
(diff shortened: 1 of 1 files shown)


Achloryn:
I'm kinda confused by this. Now this won't send me down the Disembowl>Chaos Thrust line until Chaotic Spring has completely fallen off of the target. Usually once I cycle back around to that ability, it has 2-3 seconds left, so it instead sends me down the Vorpal Thrust line, even though it would be off by the time I hit Chaos Thrust again.

Akechi-kun:
You will never maintain 100% uptime with Chaotic Spring. 2.5sks (which is standard & only SkS supported for DRG) causes Chaotic Spring to drop for roughly a second before it's reapplied, this is due to 2.5 having a 10 GCD combo, so 25s in-between, but the DOT duration is 24s. This is how the code was previously was before I borked this line & is still recommended by The Balance to do so.

Achloryn:
> You will never maintain 100% uptime with Chaotic Spring. 2.5sks (which is standard & only SkS supported for DRG) causes Chaotic Spring to drop for roughly a second before it's reapplied, this is due to 2.5 having a 10 GCD combo, so 25s in-between, but the DOT duration is 24s. This is how the code was previously was before I borked this line & is still recommended by The Balance to do so.

Sorry, let me rephrase, because I think I didn't get my point across. The issue isn't not having 100% uptime on chaotic spring. The issue is ok, so the rotation (not counting any OGCD buttons) is: 
true thrust > spiral blow > chaotic spring >wheeling thrust > drakesbane >raiden thrust > lance barrage > heaven's thrust > fang and claw > drakesbane > raiden thrust -- Here, where I should go into Spiral Blow again, it is instead sending me into Lance Barrage again because there's still 2-3 seconds of Chaotic Spring uptime. I have to actually wait until Chaotic Spring drops completely before I get the option to use Spiral Blow.

FWIW, This is all without food, and no materia melded on my gear at all, let alone intentionally statting out for SkS, but i'm still at 2.45 just from the tomestone gear mostly.

I hope this helps explain what I was getting at. I can also make a video on a training dummy and show you what I mean.

Akechi-kun:
If you're using any forced SkS, it should still work fine.
2.45 still meets the threshold of when to refresh the buffs.
I'm using 2.46sks DRG and I cannot replicate this issue on my PR. 
Please make sure you've pulled this correctly or are not using the live version, because after 12mins at the dummy & Chaotic Spring being reapplied perfectly everytime, I'm more than positive this is fixed.

Achloryn:
> Please make sure you've pulled this correctly or are not using the live version, because after 12mins at the dummy & Chaotic Spring being reapplied perfectly everytime, I'm more than positive this is fixed.

My apologies, I thought this was for the live version. I had it update just today on Dalamud, so I assumed this was a part of that. My mistake!

Akechi-kun:
@Kagekazu is handling the DRG issues, which includes this fix as well. Closing.

PR #1609, opened 2024-07-24
[DRK] Update Opener, fix action combos are on

### LINKED ISSUE #1136: [DRK] AOE combo replacements
Hello, I'm just simply wondering why the AOE combo function replaces the first skill (Overpower) in the combo for WAR but the second one (Stalwart Soul) for DRK. Is there any consideration on this?

edit: My problem is that, I think it could be better if similar logics can be applied to all the 4 tank jobs

### LINKED ISSUE #1608: [DRK] Opener has changed
The new opener no longer waits 3 GCDs - this needs updated. This is the new opener:

xX1ZZmS

zbee:
- [X] Update Opener to 1st GCD (fixes #1608)
- [X] Move both combos to the first action in the chain (fixes #1136)
- [X] ~~Implement strict JobHelper Opener with option~~ Unnecessary
- [X] Reduced the possible ranges for Living Dead's thresholds, added stepping to them - both to make the sliders slightly easier to use
- [X] Moved to `UserInt`s directly, with suggested defaults

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1006,10 +1006,10 @@ namespace XIVSlothCombo.Combos
 
         #region DARK KNIGHT
 
-        #region Souleater (Single Target) Combo
+        #region Advanced Single Target Combo
 
-        [ReplaceSkill(DRK.Souleater)]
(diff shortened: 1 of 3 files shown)


zbee:
>[!NOTE]
> DRK's Combos have moved from `Souleater` and `Stalwart Soul` **> to >** `Hard Slash` and `Unleash`.
> All hotbars will need updated.

PR #1619, opened 2024-07-25
WHM PoM Weave choice Checkbox for aoe

edewen:
I added a check box to choose whether to only weave PoM on Aoe combo

Because 

Holy spam doesn't give room for a weave so you need to manually force it with movement or wait for rapture overcap protection. The dps gain to start cranking out glare 4 and keep your 2 min cd rolling in dungeons would far outweigh the tiny gcd delay.


edewen:
Closing this. I learned about branches and with an ast change coming i wanna split them up. It will be readded.

PR #1637, opened 2024-07-28
GNB Opener Complete

EYEZOexe:
- Added Simple Single Target Option It will automatically do the opener

- Added Advanced Single Target Option user can choose if they want the level 100 opener

### DIFF
--- a/.idea/.idea.XIVSlothCombo/.idea/.name
+++ b/.idea/.idea.XIVSlothCombo/.idea/.name
@@ -1,0 +1,1 @@
+XIVSlothCombo
--- a/.idea/.idea.XIVSlothCombo/.idea/indexLayout.xml
+++ b/.idea/.idea.XIVSlothCombo/.idea/indexLayout.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="UserContentModel">
+    <attachedFolders />
+    <explicitIncludes />
+    <explicitExcludes />
+  </component>
+</project>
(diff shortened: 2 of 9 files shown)


Kagekazu:
if im not mistaken u have removed everything except the opener?

EYEZOexe:
I've commented it out to make it more clear what the changes are and how its implemented

EYEZOexe:
I am planning on doing a complete GNB revamp since its currently bricking in bloodfest due to cartridge logic

Kagekazu:
true, but as it is now, if we merge, it will comment out everything but the opener. Please keep it uncommented.
Git itself shows the changes already :)

Also make sure it doesnt conflict with PR #1600

EYEZOexe:
ill have a look once again in that case, thanks for your time

.idea/.idea.XIVSlothCombo/.idea/.name
zbee, line 1: Your IDE settings should not be included in this PR.

Akechi-kun:
This job is not that deep to where it _needs_ to have its opener added into JobHelpers. We essentially do the same burst every 60s aside from certain skills being moved around the burst because there's no existent CD (i.e. Sonic Break), so in my humblest opinion, this isn't necessary. 

The opener you added into **JobHelpers** is only for the scenario of "_0 carts with Bloodfest_", translating to "_Brutal Shell (NoMercy, Bloodfest) > Burst_", which is only one of many different scenarios you will face when bursting. (i.e. _2 cart without Bloodfest_? _3 cart without Bloodfest_? _2 cart with Bloodfest for **Lv100** only_? _3 cart with Bloodfest for **Lv90** only_?)

Furthermore, This opener will _only_ work for Lv100 (I assume this is just a WIP). Keep in mind if you are going to "revamp" GNB, you need to take into account every scenario GNBs will face, and not just at Lv100, but from Lv1 - Lv100.

The logic in the live version is currently weak, so in #1600 I added logic that is much stronger & more than enough to suffice for every scenario. Feel free to take a look for yourself. 👍

EYEZOexe:
didnt see this, thanks for the heads up

PR #1646, opened 2024-07-30
[GNB] Added Lvl 100 Balance Opener

EYEZOexe:
image

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -1428,61 +1428,67 @@ namespace XIVSlothCombo.Combos
         [ReplaceSkill(GNB.KeenEdge)]
         [CustomComboInfo("Advanced Gunbreaker Feature", "Replace Keen Edge with its combo chain and uses Burst Strike to prevent ammo overcap. **2.5sks ONLY**", GNB.JobID)]
         GNB_ST_MainCombo = 7001,
+        
+        #region Level 100 Balance Opener
+        [ParentCombo(GNB_ST_MainCombo)]
+        [CustomComboInfo("Level 100 Balance Opener", "Adds the Balance opener for 2.50GCD to the rotation.", GNB.JobID)]
+        GNB_ST_Opener = 7002,
+        #endregion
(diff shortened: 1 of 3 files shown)


XIVSlothCombo/Combos/CustomComboPreset.cs
zbee, line 1: Changing option IDs like this will break users' setups.

zbee (commented):
At least in terms of the user config, I don't think this is ready.

XIVSlothCombo/Combos/CustomComboPreset.cs
EYEZOexe: how can I add another feature then without changing the other IDs? do I just pick a random one? I thought the ID's were the ordering

EYEZOexe: Nvm fixed

EYEZOexe:
image

XIVSlothCombo/Combos/CustomComboPreset.cs
zbee: The next highest available ID within the same thousands block.

No, you can order with the Custom Attribute:
```cs
[CustomComboInfo("Feature", "Description", GNB.JobID, 1)]
//                                                    ^ this is the order
```
You can refer to [DNC](https://github.com/Nik-Potokar/XIVSlothCombo/blob/5a14682a4bbecb52378902fc34efcaffbf9b7bee/XIVSlothCombo/Combos/CustomComboPreset.cs#L868).

EYEZOexe: gotcha ill use that for next time in that case! is it good?

zbee: If it appears at the top of the list as it is now, I think the PR is fine now from the perspective of `CustomComboPreset`.
I'm not a GNB, so I won't speak to the rest.

EYEZOexe: I used RPR as an example

image

and now the ordering and position of code in `CustomComboPreset`

PR #1654, opened 2024-08-01
[VPR] Vicewinder Dynamic First Coil Selection

memorial36b:
Adds a feature to dynamically select whether Hunter or Swiftskin's Coil should be used after Vicewinder based on current position, defaulting to the selected Rear/Flank First option if player is not on rear or flank.
Option added to Advanced ST and standalone Vicewinder combo feature -- simple mode untouched.


memorial36b:
feature added to #1651, this is now redundant

PR #1662, opened 2024-08-03
[WHM] Prioritize Raise in Swiftcast>Raise Feature

dstormcaller:
ActionReady checks Swiftcast instead of checking whether the buff is active. This is so that Raise is still presented after Swiftcast buff is used/lost, in case a hardcast raise is necessary.


Taurenkey:
This would be incorrect, as this is supposed to be Swiftcast -> Raise, and not Raise -> Swiftcast. That option is the global healer raise feature, and this is the alternative version that swaps it about.

PR #1675, opened 2024-08-06
[BRD] Change Jaw When  RagingStrikes buff remaining 4sec

Redmoonwow:
I added Bard's DOT update because some of the hardcore players are involved.


Redmoonwow:
Temporarily closed to allow for new collection

PR #1698, opened 2024-08-10
[WAR] fixes/adjustments

Akechi-kun:
**Changelog:**
- [x] add _Simple ST & AoE_ rotations for newcomers 
- [X] fix issue with Primal Rend settings being ignored... again
- [x] last minute prep before merge to **v3.2.0.7**

### DIFF
--- a/XIVSlothCombo/Combos/PvE/WAR.cs
+++ b/XIVSlothCombo/Combos/PvE/WAR.cs
@@ -71,14 +71,92 @@ namespace XIVSlothCombo.Combos.PvE
                 WAR_InfuriateAoEGauge = "WAR_InfuriateAoEGauge";
         }
 
-        // Replace Storm's Path with Storm's Path combo and overcap feature on main combo to fellcleave
-        internal class WAR_ST_StormsPath : CustomCombo
+        internal class WAR_ST_SimpleMode : CustomCombo
         {
-            protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.WAR_ST_StormsPath;
+            protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.WAR_ST_Simple;
 
             protected override uint Invoke(uint actionID, uint lastComboMove, float comboTime, byte level)
             {
-                if (IsEnabled(CustomComboPreset.WAR_ST_StormsPath) && actionID == StormsPath)
+                if (actionID == StormsPath)
+                {
+                    var gauge = GetJobGauge<WARGauge>().BeastGauge;
+                    float GCD = GetCooldown(HeavySwing).CooldownTotal;
+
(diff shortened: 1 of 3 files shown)


XIVSlothCombo/Combos/PvE/WAR.cs
```diff
@@ -352,7 +479,7 @@ protected override uint Invoke(uint actionID, uint lastComboActionID, float comb
                 {
                     if (LevelChecked(PrimalRend) && HasEffect(Buffs.PrimalRendReady))
                         return PrimalRend;
-                    if (LevelChecked(PrimalRuination) && HasEffect(Buffs.PrimalRuinationReady) && WasLastWeaponskill(PrimalRend))
+                    if (LevelChecked(PrimalRuination) && HasEffect(Buffs.PrimalRuinationReady) && JustUsed(PrimalRend))
```
Gakai, line 482: I think, `JustUsed` is unnecessary (and potentially wrong) here.
Unnecessary because if the level is ok and you have the buff, then it should be returned. Actually the `HasEffect` check alone would be sufficient, I think, as you cannot get the buff on lower levels :)

Maybe wrong because Primal Rend is a normal GCD, and if I understand `JustUsed` correctly, it checks first if Primal Rend is on cooldown, which is false, when GCDs (e.g. Primal Ruination) are usable.

Kagekazu: > I think, `JustUsed` is unnecessary (and potentially wrong) here. Unnecessary because if the level is ok and you have the buff, then it should be returned. Actually the `HasEffect` check alone would be sufficient, I think, as you cannot get the buff on lower levels :)
> 
> Maybe wrong because Primal Rend is a normal GCD, and if I understand `JustUsed` correctly, it checks first if Primal Rend is on cooldown, which is false, when GCDs (e.g. Primal Ruination) are usable.

justused checks the last actionid that has been used, which in this case would be Primal Rend

Gakai: Thanks, but I still think, it is incorrect, because it does not HAVE to be the action right before Primal Ruination. You can use any other GCD between them, as long as the buff is present.

Akechi-kun: You are correct @Kagekazu, however this is only base _JustUsed_, so it's only checking if **PrimalRend** was used for the last 3 seconds, and I initially thought you _had_ to press **Ruination** right after **Rend**, and this is also for the "Primal Combo Feature", which only replaces **InnerRelease** with **PrimalRend** & **PrimalRuination**. _JustUsed_ would be more valuable inside the one-button as we usually _want_ to use **Rend>Ruination** asap, but here the user wants to hold their certain Primal Combo actions, so only have the _HasEffect(Buffs.PrimalRuinationReady)_ should suffice for this specific class module.

Thanks for the feedback, will edit this.

Issue #1706, opened 2024-08-11
[SAM] Kaeshi: Settsugekka not being spent


kazumaKiryu542:
Had a few times where it looks like Kaeshi: Settsugekka isn't being used after Midare: Settsugekka. Few different instances, all at Lv100.

PR #1711, opened 2024-08-11
[SAM] Standalone Combo Adjustments

### LINKED ISSUE #1704: [REQUEST] SAM Kasha & Yukikaze Option
At the moment, while using **Advanced ST option** and turning on **Kasha & Yukikaze** combo will prevent the other individual Yukikaze & Kasha combo to be deactivated due to conflict. 

Would appreciate if they would not and be able to be on simultaneously alongside the Advanced ST Option. 

At the same time, if it would be possible to **allow for a combo string that would be changed into Gekko** (perhaps for example on the Jinpu or Gyofu/Hakaze button). 

These option would be useful especially when micro optimizing to avoid using a higabana application where it is not necessary.

Perhaps there could also be a way to make **Gyofu/Hakaze a button that would only perform filler GCD-only**?

Taurenkey:
- [x] All 3 single target combo chains have their own features (was previously just Yukaze & Kasha, now includes Gekko).
- [x] Both AoE combo chains have their own features (was previously just Oka, now includes Mangetsu).
- [x] Kenki overcap options now separated per feature and is now optional for the standalone combo chains.
- [x] Move `Simple/Advanced Single Target` action being overriden from Gekko to Hakaze/Gyofu to remove the need for conflicting combo tag.
- [x] Removed conflicting combos throughout the job.
- [ ] Update names/descriptions to be more in line with established formatting.
- [ ] #1706.
- [X] Closes #1704 
- [X] Higabana now simplified to be applied more consistently for both `Simple & Advanced Single Target`.
- [X] Meikyo Shisui logic adjusted to not start mid-combo and used only before fully capping charges.
- [X] Senei & Ogi Namikiri will now fire off more consistently.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3201,33 +3201,35 @@ namespace XIVSlothCombo.Combos
 
         #region SAMURAI
 
-        #region Yukikaze/Kasha Combos
+        #region Yukikaze/Kasha/Gekko Combos
 
         [ReplaceSkill(SAM.Yukikaze)]
-        [ConflictingCombos(SAM_ST_Yukikaze)]
         [CustomComboInfo("Yukikaze Combo", "Replace Yukikaze with its combo chain.", SAM.JobID)]
         SAM_ST_YukikazeCombo = 15000,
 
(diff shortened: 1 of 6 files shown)


mahfujkadir:
it be good for sam to get guren into the lvl 70 rotation since you cant get senei for lvl 70 content

PR #1712, opened 2024-08-11
[SGE] Hold Phlegma stack option

Jukkales:
Option to hold Phlegma and not use it on cooldown - does not affect opener option and has overcapping protection.

grafik

### DIFF
--- a/XIVSlothCombo/Combos/PvE/SGE.cs
+++ b/XIVSlothCombo/Combos/PvE/SGE.cs
@@ -110,7 +110,8 @@ namespace XIVSlothCombo.Combos.PvE
             #region DPS
             public static UserBool
                 SGE_ST_DPS_Adv = new("SGE_ST_DPS_Adv"),
-                SGE_ST_DPS_EDosis_Adv = new("SGE_ST_Dosis_EDosis_Adv");
+                SGE_ST_DPS_EDosis_Adv = new("SGE_ST_Dosis_EDosis_Adv"),
+                SGE_ST_DPS_Phlegma_Adv = new("SGE_ST_DPS_Phlegma_Adv");
             public static UserBoolArray
                 SGE_ST_DPS_Movement = new("SGE_ST_DPS_Movement");
             public static UserInt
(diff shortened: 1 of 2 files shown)


XIVSlothCombo/Combos/PvE/SGE.cs
```diff
@@ -416,7 +418,18 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
...
                             uint phlegma = OriginalHook(Phlegma);
-                            if (InActionRange(phlegma) && ActionReady(phlegma)) return phlegma;
+
+                            if (InActionRange(Phlegma) && ActionReady(Phlegma))
+                            {
+                                if (!Config.SGE_ST_DPS_Phlegma_Adv || GetRemainingCharges(Phlegma) == GetMaxCharges(Phlegma)) 
+                                    return phlegma;
+                                
+                                if (GetRemainingCharges(Phlegma) == 1 && 
+                                    GetCooldownChargeRemainingTime(Phlegma) <= Config.SGE_ST_DPS_PhlegmaThreshold)
```
zbee, line 428: Does not check `Config.SGE_ST_DPS_Phlegma_Adv` only a sub-option of it.

XIVSlothCombo/Combos/PvE/SGE.cs
```diff
@@ -416,7 +418,18 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
...
+                                {
+                                    return phlegma;
+                                }
+                            }
```
zbee, line 432: Would prefer to see this all rolled into a single `if`, a la `Eukrasia`.

zbee (commented):
Please tighten up the conditions around this.

XIVSlothCombo/Combos/PvE/SGE.cs
Jukkales: its the else branch of the adv-config option (or full stacks). instead of two ifs it could be if-else true

PR #1715, opened 2024-08-12
[PCT] Lvl 100 Opener, Advanced ST, Advanced AoE 

AkiNeeChan:
- [X] Added Lvl 100 PCT Balance opener
- [X] Added all in one lower Lvl openers 92, 90, 80, 70 
- [X] Added early Starry Muse choice to the openers
- [X] Added Advanced ST option with customization options
- [X] Added Advanced AoE option with customization options
- [X] Add levelchecks
- [X] Fixed mog/madeen not working on AoE - AoE just blasts now
- [X] Add additional options to expand customization + Few tiny fixes

Seems good to go! 
Few things i wanna update later on is re-working the opener code to be all in one piece and not segmented as it is now.
Plus other optimizations to come when i have time to do more testing.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -2366,6 +2366,10 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Balance Opener Option", $"Uses the Balance Opener depending on your current level. \n Supports lvl 70, 80, 90, 92, 100. \n - Requirements: \n - Starry Muse off cooldown \n - Pom. Weapon, Landscape Motif", PCT.JobID)]
         PCT_ST_Advanced_Openers = 20006,
 
+        [ParentCombo(PCT_ST_Advanced_Openers)]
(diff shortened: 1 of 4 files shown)


FreeAura:
Hey, I know this isn't fully ready based on the checklist, and I'm not sure if you're open here to feedback but just thought i'd comment after giving it a whirl with its current state:

Rainbow drip currently sometimes (most of the time) doesn't fit into the starry muse window below lvl 100 > currently testing at 97, it's really really tight.

Madeen/Mog of the Ages is also sometimes not under buffs/starry muse

It currently tries to weave motif abilities under the normal 1.5 gcd, which often causes clipping and weaving problems

If I find any bugs or have any suggestions etc, I'm more than happy to report here if you're okay with it. Thanks for working on Picto! <3

Nik-Potokar:
Hey! Thanks for the feedback!

The burst window for Starry Muse is quite tight, as it requires fitting in 2x comets or 1x comet along with a GCD, Starprism, Hammer combo, and Rainbowdrip. As a result, Rainbowdrip will almost always be snapshotted by the last second or buff tick of Starry Muse. 

I'll also review Mog and Madeen over the weekend since they still need improvement.

Lastly, I'll take a closer look at the clipping issues.

If you have any Xivanalysis timelines please DM them to me so i can have a look.

PR #1747, opened 2024-08-23
[MNK] fixes/adjustments

Akechi-kun:
- [x] Fix issue with _Brotherhood_ option in _Advanced Mode - Single Target_ module being ignored
- [x] Fix issue with _RiddleOfFire_ option in _Advanced Mode - Single Target_ module being ignored

### DIFF
--- a/XIVSlothCombo/Combos/PvE/MNK.cs
+++ b/XIVSlothCombo/Combos/PvE/MNK.cs
@@ -294,13 +294,15 @@ namespace XIVSlothCombo.Combos.PvE
 
                         if (IsEnabled(CustomComboPreset.MNK_STUseBuffs))
                         {
-                            if (Brotherhood.LevelChecked()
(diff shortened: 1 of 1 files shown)


Akechi-kun:
@sammhill if you plan to do a PR like this of your own, lmk & I will merge this into yours or just close.

kiramon2:
In Simple AoE, it will use Perfect Balance -but it wont move past. Not sure if this can be stated here

Akechi-kun:
> -but it wont move past

Need specifics please.

Indygoace:
I did notice in the DT Monk, Brotherhood is pushed ahead of Riddle of Fire. The issue I noticed so far is in both advanced and in Simple melee (and in AOE) Perfect Balance is not an OGCD it is forced to do that rotation, which causes the combo to freeze.

Also the on the (1mins) | (3mins) | (5mins) The combo tries to force 2 Perfect balance rotations on the [Solar Lunar] option (I don't know if that is intended though.

PR #1750, opened 2024-08-23
[PCT] Simple Modes, Motif % stop in advanced

### LINKED ISSUE #1768: [PCT] Request: Have percentage limit for casting motifs
It's frustrating when something is almost dead, and you're casting motifs instead of doing damage (especially since you'll be able to instant cast them out of combat). Would be nice to have an option like SAM does for Higanbana where it won't use it when the target is below a certain HP percentage.

edewen:
Copied the ST advanced mode and removed configuration requirements for an optimized Simple mode
Fixes #1768

- [x] ST Simple mode Optimized based on advanced
- [x] Aoe Simple mode Optimized based on advanced
- [x] Holy in white added to simple aoe. Retains one charge for movement/comet
- [x] Added slider for each motif in advanced st and aoe to stop drawing. Done separately in case some are trying to save the 2min but not the hammer/moogle
- [x] Rainbow drip as moving option higher priority than Holy in white

### DIFF
--- a/XIVSlothCombo/Combos/PvE/PCT.cs
+++ b/XIVSlothCombo/Combos/PvE/PCT.cs
@@ -87,95 +87,248 @@ namespace XIVSlothCombo.Combos.PvE
         internal class PCT_ST_SimpleMode : CustomCombo
         {
             protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.PCT_ST_SimpleMode;
+            internal static PCTOpenerLogicLvl100 PCTOpenerLvl100 = new();
+            internal static PCTOpenerLogicLvl92 PCTOpenerLvl92 = new();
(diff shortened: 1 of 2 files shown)


Akechi-kun:
While you're at it, add in _Rainbow Drip_ under the movement options. I've been noticing a slight issue with _Rainbow Drip_ being lower priority than _Holy In White_ when moving, and it's causing _Drip_ to fall out of _StarryMuse_ if moving.

Akechi-kun (commented):
> Added slider for each motif in advanced st and aoe to stop drawing. Done separately in case some are trying to save the 2min but not the hammer/moogle

This wasn't added into **CustomComboPreset.cs**, so it's not visible in the PCT menu.

edewen:
not sure what i missed on second comment but... 

rainbow drip on moving heard chef!

PR #1759, opened 2024-08-24
[AST] Aspected helios hot check fix

edewen:
Fixed the missed buff name and removed horoscope forcing aspected. any helios works for horoscope.

Closes discord report


Taurenkey:
Already done, woops lol

PR #1760, opened 2024-08-25
[AST] Earthly Star Priority Shift

edewen:
Reverted _Added earthly star to the prepull noncombat line with the card draw so you can plant your star before pull_ 

Shifted earthly star a bit higher above the ogcds that cant drift. oracle/arcana
Added tag to not be moving to lay earthly star in aoe to help prevent it wasting during aoe pulls
Added a buff check to the opener for lightspeed to fix rare double cast (discord rpt)

### DIFF
--- a/XIVSlothCombo/Combos/PvE/AST.cs
+++ b/XIVSlothCombo/Combos/PvE/AST.cs
@@ -452,7 +452,7 @@ namespace XIVSlothCombo.Combos.PvE
                         ActionWatching.NumberOfGcdsUsed >= 3)
                         return Divination;
                     //Earthly Star
-                    if (IsEnabled(CustomComboPreset.AST_AOE_DPS_EarthlyStar) &&
+                    if (IsEnabled(CustomComboPreset.AST_AOE_DPS_EarthlyStar) && !IsMoving &&
                         ActionReady(EarthlyStar) &&
                         CanSpellWeave(actionID))
                         return EarthlyStar;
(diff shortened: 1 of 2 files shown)


XIVSlothCombo/Combos/PvE/AST.cs
```diff
@@ -215,11 +215,15 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
...
                     !InCombat()))
-
+                {
                     if (IsEnabled(CustomComboPreset.AST_DPS_AutoDraw) &&
                         ActionReady(OriginalHook(AstralDraw)) && (Gauge.DrawnCards.All(x => x is CardType.NONE) || (DrawnCard == CardType.NONE && Config.AST_ST_DPS_OverwriteCards)))
                         return OriginalHook(AstralDraw);
 
+                    if (IsEnabled(CustomComboPreset.AST_ST_DPS_EarthlyStar) &&
+                        ActionReady(EarthlyStar))
+                        return EarthlyStar;
```
zbee, line 225: In my opinion this should be a sub-option of the Opener, primarily so we can note in the description that it should target your target, not your self.
> To be used in conjunction with Redirect/Reaction/etc set up to place Earthly Star at your `<target>`.

As it is now, you can have Reaction/Redirect/etc place Earthly Star under your `<focus>` or `<self>` no problem, but theoretically if it's in the opener it could not reach the boss with that setup (e.g. Alliance Raids).

---

As a sidenote, this note is why I did not add Star to the Opener when I added Earthly Star: I want to use `<self>` so that the star's healing effects are not lost when used on wall bosses (e.g. Endsinger, your Star can cover most of the arena, but not if placed at `<target>`).
I am not opposed to its addition at all, but in my opinion it MUST be noted that it may not work without `<target>` being your setting.

PR #1762, opened 2024-08-25
[PCT] Add AoE Holy in White feature

memorial36b:
As per [The Balance FAQ](https://www.thebalanceffxiv.com/jobs/casters/pictomancer/faq/#q2), Holy is a DPS gain in AoE. Adds a feature to include Holy in White to the AoE combo and a slider to choose how many charges to hold on to.


edewen:
It looks like you only touched on the advanced. so that works I added it to the simple modes pr as well saving a single charge for aoe.

PR #1763, opened 2024-08-25
[VPR] Add options to disable range checks in AoE combo

memorial36b:
Removes the action range checks from the Vicepit combo sequence and Reawaken in the AoE combo (simple and advanced), so they will still fire without a target selected.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -4148,7 +4148,7 @@ namespace XIVSlothCombo.Combos
         VPR_AoE_Vicepit = 30105,
 
         [ParentCombo(VPR_AoE_Vicepit)]
-        [CustomComboInfo("Disable Range Check", "Disables the range check for Vicepit.", VPR.JobID)]
+        [CustomComboInfo("Disable Range Check", "Disables the range check for Vicepit, so it will be used even without a target selected.", VPR.JobID)]
         VPR_AoE_Vicepit_DisableRange = 30111,
 
         [ParentCombo(VPR_AoE_CDs)]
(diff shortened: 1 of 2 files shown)


Akechi-kun:
Why, if I may ask? 

This could cause issues for users who mindlessly hold the one-button without checking which skills are being used. 

_Vicepit_ & it's combo route _should_ only be used when in range of a target, even if you have to hold one coil from one mob to another as it's better to do this than to just waste the skill & damage altogether because you weren't within range. 

This is also notably bad for _Reawaken_, since you _want_ to be in range of the target to get the 750 potency from the hit regardless of any circumstance.

I do not recommend this PR.

memorial36b:
I have added options to disable the range checks. The only changes I have made to the logic itself is that I have changed the melee range checks to 5y range checks, as I believe that is fair given the Vicepit combo's 5y radius.

I would argue that at minimum the Vicepit combo _excluding Vicepit itself_ should have its range checks removed, as currently if you lose your target while in the middle of the Vicepit combo (ie a dungeon trash pack target dying mid-combo) it reverts to the filler combo, wasting the rest of the combo even though other enemies may still be within AoE range. By retaining the range check for Vicepit itself, the opportunity to waste the combo without enemies in range is still mitigated.

Akechi-kun (commented):
👍🏽💯

Akechi-kun:
> currently if you lose your target while in the middle of the Vicepit combo (ie a dungeon trash pack target dying mid-combo) it reverts to the filler combo, wasting the rest of the combo even though other enemies may still be within AoE range if you don't realize what has occurred and continue pressing the button.

I see, didn't know this was a thing. I suppose these option additions would help as a band-aid fix for that specifically for now, but I'll do some more digging on that scenario.

PR #1770, opened 2024-08-27
Function cleanup.

Tartarga:
Removal of function middlemans in ActionWatching. Various functions will call ActionWaiting directly in Action.cs (cleaning up my crap...) 
ActionRange moved to Target.cs
Condensed two OutOfRange functions to one.
Removed duplicate GetMaxCharges

JobHelpers/BLU.cs BLUHelper added.
ActionWatching: Moved GetBLUIndex to BLUHelper
Presets: Updated BLUHelper reference
Debug: Two Null Warnings squashed


Tartarga:
Coming back to later

Issue #1773, opened 2024-08-27
[SGE] Psyche not working properly on single target


ExoLeviticos:
Psyche is listed as "Psych option - Weaves Psych when available." under the single target DPS feature. This option currently doesn't work properly.

Under movement options, there are four checkmarks to be selected (Toxikon, Dyskrasia, Eukrasia and Psyche), and that option does work properly, as Psyche is added to the combo during movement, even despite it not being a proper movement tool, considering its a oGCD.
The AoE option also works properly.

I assume it the typo "Psych" on the single target option has something to do with it not working.

edewen:
That typo isn't actually in the code part so that's a non issue there. So when you are just using the st button. it is not psyching in the opener/rotation? As i am unable to replicate, it weaves it on cd for me. perhaps uncheck and check it all?

as for the movement, yeah its ogcd i agree it shouldn't be there. Trying the fix your actual issue first

Kagekazu:
FWIW, Psyche does weave properly on ST for me.

Charles-YYH:
Hello, I have the same issue for Psyche. 
Weird thing is that It's working correctly during two GCDs window (for example, Phlegma ) but it does not work during one GCD window (Dosis or Eukrasian).

Charles-YYH:
> Hello, I have the same issue for Psyche. Weird thing is that It's working correctly during two GCDs window (for example, Phlegma ) but it does not work during one GCD window (Dosis or Eukrasian).

Update, this issue has been resolved after disable and enable this option.

PR #1774, opened 2024-08-27
[SGE] Psyche Removal from movement, it's ogcd

edewen:
- [x] Removed psyche from movement, it is an ogcd.
- [x] Fixed typo in custom-combo, irrelevant but may as well.
- [x] Moved psyche up in priority as i noticed a lucid dreaming caused it to drift. might help #1773 but we cant replicate

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3256,7 +3256,7 @@ namespace XIVSlothCombo.Combos
         SGE_ST_DPS_Rhizo = 14007,
 
         [ParentCombo(SGE_ST_DPS)]
-        [CustomComboInfo("Psych Option", "Weaves Psych when available.", SGE.JobID, 112, "", "")]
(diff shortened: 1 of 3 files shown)


Taurenkey:
What's the reason for removing it from movement? Generally that's where we'd put oGCDs since they don't have cast times so are suited for uptime during movement.

edewen:
1. sage has gcd movement options 100% of the time. toxicon, eukrasia, and dyskrasia. It will even use phlegma on its own as movement before it gets to movement based on the movement code being at the bottom. 

2. On that same note. Psyche is higher in priority outside of movement as an ogcd. So it will always end up used before movement which would be confusing to those expecting it for movement. 

3. That is not where we would put off global cooldowns. Thats where we would put instant casts. Not a single healer supports that statement. sch ruin 2, ast lightspeed bc it is designed as a movement, and whm doesn't have anything at all.

Taurenkey:
Rejecting as it isn't beneficial, it's stripping away valid options.

PR #1793, opened 2024-09-04
[DRK] Fix Delirium Bloodspiller usage under 96

### LINKED ISSUE #1764: [DRK] the DRK_ST_Bloodspiller option only accounts for one of two Delirium buffs.
It accounts for the 3836 buff, but before lvl 96 the buff id is actually 1972. So the feature doesn't work until you're level 96.

zbee:
- [X] Add older Delirium buff ID, and check it when below the level of Enhanced Delirium when trying to use free Bloodspiller procs.

Closes #1764

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRK.cs
+++ b/XIVSlothCombo/Combos/PvE/DRK.cs
@@ -59,8 +59,8 @@ namespace XIVSlothCombo.Combos.PvE
             public const ushort
                 // Main Buffs
                 BloodWeapon = 742,
-                OlderDelirium = 1972,
-                Delirium = 3836,
+                Delirium = 1972,
+                EnhancedDelirium = 3836,
 
                 // Periodic Buffs
                 Darkside = 741,
@@ -237,7 +237,7 @@ namespace XIVSlothCombo.Combos.PvE
                 if (LevelChecked(Delirium)
                     && LevelChecked(ScarletDelirium)
                     && IsEnabled(CustomComboPreset.DRK_ST_Delirium_Chain)
-                    && HasEffect(Buffs.Delirium)
+                    && HasEffect(Buffs.EnhancedDelirium)
                     && gauge.DarksideTimeRemaining > 0)
                     return OriginalHook(Bloodspiller);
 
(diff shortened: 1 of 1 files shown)


XIVSlothCombo/Combos/PvE/DRK.cs
```diff
@@ -59,6 +59,7 @@ public static class Buffs
             public const ushort
                 // Main Buffs
                 BloodWeapon = 742,
+                OlderDelirium = 1972,
```
Tartarga, line 62: Just a suggestion,
Would Delirium 1972 and ScarletDelirium 3836 be better names?

XIVSlothCombo/Combos/PvE/DRK.cs
```diff
@@ -244,16 +245,16 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                 if (LevelChecked(Delirium)
                     && IsEnabled(CustomComboPreset.DRK_ST_Bloodspiller))
                 {
-                    //Regular Bloodspiller
-                    if (GetBuffStacks(Buffs.Delirium) > 0)
+                    //Bloodspiller under Delirium
+                    if (GetBuffStacks(Buffs.Delirium) > 0
+                        || (TraitLevelChecked(Traits.EnhancedDelirium)
+                            && GetBuffStacks(Buffs.OlderDelirium) > 0))
                         return Bloodspiller;
```
Tartarga, line 252: I'm not a DRK so I'm trying to follow.
I'm reading "If Scarlet Ready Delirium > 0 OR [ We meet level to do scarlet delirium checking but checking if older is > 0 ] " Looks like you got your buffs switched?

Also, what about
```
ushort DeliriumBuff = !TraitLevelChecked(Traits.EnhancedDelirium) ? Delirium : ScarletDelirium;
If GetBuffStacks(DeliriumBuff) > 0 
```

XIVSlothCombo/Combos/PvE/DRK.cs
zbee: Scarlet Delirium would get me confused with the ability of the same name, I went for "Enhanced" in [b2b867a](https://github.com/Nik-Potokar/XIVSlothCombo/pull/1793/commits/b2b867a96a4159c0b762f82f949bb1be59b86ec9).

XIVSlothCombo/Combos/PvE/DRK.cs
zbee: Great catch, I forgot to restore an exclamtion mark to flip the logic after testing.
And great suggestion for making the logic a bit easier to read.
This is implemented in [b8f3b5f](https://github.com/Nik-Potokar/XIVSlothCombo/pull/1793/commits/b8f3b5f54bac07b1395ba06ebf16a1757edc8679).

PR #1810, opened 2024-09-15
[SCH] Aetherflow Fix

edewen:
- [x] Added line to aetherflow to not use if dissipation was just used. Bad server tick causing it not to register the aetherflow stack in time likely due to animation lock skipping. reported via combo issues.


Kagekazu:
issue : https://discord.com/channels/1001823907193552978/1284756774200803433

PR #1814, opened 2024-09-18
[SGE] Sage Pneuma to zoe downleveled fix

edewen:
image

- [x] Added or statement so below 90, Pnuema shows Zoe. No more need to put it somewhere else just for downleveled content.

### DIFF
--- a/XIVSlothCombo/Combos/CustomComboPreset.cs
+++ b/XIVSlothCombo/Combos/CustomComboPreset.cs
@@ -3495,8 +3495,8 @@ namespace XIVSlothCombo.Combos
         [CustomComboInfo("Taurochole to Druochole Feature", "Turns Taurochole to Druochole when Taurochole is on cooldown.", SGE.JobID, 700, "", "")]
         SGE_TauroDruo = 14038,
 
-        [ReplaceSkill(SGE.Pneuma)]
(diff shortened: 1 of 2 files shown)


Akechi-kun:
I would consider reworking this whole option altogether. 

You're doing double-work by putting a low-level skill onto a high-level skill, when really you can just change this option's logic to be on **Zoe** instead of **Pneuma**.

Akechi-kun:
I would consider reworking this whole option altogether. 

You're doing double-work by putting a low-level skill onto a high-level skill, when really you can just change this option's logic to be on **Zoe** _(Lv56)_ instead of **Pneuma** _(Lv90)._ 

For example: change the _actionID_ to **Zoe**, add the option for **Pneuma** when off cooldown & level-check it.

edewen:
The only thing I did was essentially a level check fix to not screw people over in downleveled dungeons. Not a change in functionality. 

My thinking. If you have zoe up. you always want it before pnuema. However, you may not always want pnuema after zoe. Some people may still want zoe elsewhere. 

Also wouldn't that mess with st heal option changing zoe topnuema if you also have st heal button popping up zoe?

Taurenkey:
My 2 cents on this is that it should be a seperate feature. The original intention of this feature wasn't a strict "Zoe & Pneuma one button combo" but more "Pneuma, but also Zoe when on cooldown so it's not a dead button". It's a pretty fundamental change to the feature swapping the replaced action, because it'll lock you out of Zoe when Pneuma is up, whereas the current version does not.

PR #1821, opened 2024-09-22
[RPR] Add Sacrificium only option 

### LINKED ISSUE #1820: RPR Gluttony on Bloodstalk Not also Putting Sacrificium?
There's no point in putting Gluttony on Bloodstalk if it doesn't also put Sacrificium on the button, as we still have to put Gluttony on the hotbar for that skill then. And I will not use the Enshroud Combo feature as that's too cheaty for me. I don't want one button burst. I just want to reduce button bloat.

Kagekazu:
- [x] Add Sacrificium only option to Bloodstalk/Gluttony option
- [x] Closes #1820


Kagekazu:
Merged into #1836

PR #1830, opened 2024-10-14
[DRG] Fix Wheeling Thrust & Fang and Claw

Noodlefu:
Fixed rotation logic to ensure both Wheeling Thrust & Fang and Claw fire if available. Looks like they were previously looking for the wrong preceding skill.

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRG.cs
+++ b/XIVSlothCombo/Combos/PvE/DRG.cs
@@ -232,7 +232,7 @@ internal class DRG
                         return OriginalHook(ChaosThrust);
                     }
 
-                    if (lastComboMove is ChaosThrust && LevelChecked(WheelingThrust))
+                    if (lastComboMove is ChaoticSpring && LevelChecked(WheelingThrust))
                     {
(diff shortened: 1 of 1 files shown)


Taurenkey:
I don't believe this is a fix to anything. In fact, it would lock it to only work at level 86 onwards.

XIVSlothCombo/Combos/PvE/DRG.cs
```diff
@@ -232,7 +232,7 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                         return OriginalHook(ChaosThrust);
                     }
 
-                    if (lastComboMove == OriginalHook(ChaosThrust) && LevelChecked(WheelingThrust))
```
Kagekazu, line 235: The originalhook makes it so it works for the lower levels aswell.
Changing this makes it so it only works once u got the upgraded skill

Noodlefu:
I'd misread the levels of the two combo skills, apologies - this is broken in release (the skills refuse to fire no matter what I do at level 90), but I'll take a further look and ascertain if I can't see why.

PR #1847, opened 2024-10-28
[SAM] Improve Iaijutsu and Ikishoten features

Gakai:
* Iaijutsu:
  * Handle Kaeshi: Namikiri correctly
  * Add Goken checks to Tsubame Gaeshi handling (would not apply at all previously)
  * Keep Shoha on Iaijutsu after Kaeshi use (not only if weavable)
* Kyuuten/Shinten: Add option to include Zanshin after Senei/Guren
* Ikishoten: Add option to include Shoha before Ogi Namikiri

@Akechi-kun, could you check if these adjustments are appropriate? I'm trying my hand at my first PR here :)


Gakai:
@Akechi-kun, could you check if these adjustments are appropriate?  :)

